PR #10, opened 2020-04-08
Add legalese

cevian:
Add License and Notice. Also add a blurb about contributing to README.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,177 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
mfreed, line 179: This Appendix should be cut from LICENSE, because we instead have in NOTICE.

NOTICE
```diff
@@ -0,0 +1,16 @@
+
+Timescale-Observability Helm Charts by Timescale (TM)
+
+Copyright (c) 2020-2020  Timescale, Inc. All Rights Reserved.
```
mfreed, line 4: 2020 only, not 2020-2020

PR #15, opened 2020-04-10
Add support for provisioned Grafana dashboards

blagojts:
By putting grafana dashboard json files in
a folder in the chart they can be automatically provisioned.

An example k8s monitoring dashboard is added. The dashboard
is copied from https://grafana.com/grafana/dashboards/315 
with some of the queries modified to make them up to date with 
changes in the labels


chart/values.yaml
```diff
@@ -83,4 +83,9 @@ grafana:
...
   sidecar:
     datasources:
-      enabled: true
\ No newline at end of file
+      enabled: true
+    dashboards:
+      enabled: true
+
+dashboards:
+  folder: dashboards
```
cevian, line 91: can't we put this under `grafana.sidecar.dashboards.folder`?

PR #23, opened 2020-05-04
Refactor connector queue config values

### LINKED ISSUE #21: Refactor remote_write values structure to avoid confussion with prometheus.remoteWrite
The prometheus helm chart has a `remoteWrite` config section where you can put in your configuration for the endpoints. We have a `remote_write` section for the timescale-prometheus connector that is used to manage the queue config. This causes understandable confusion

blagojts:
Resolves #21 

Previously the configuration options for Timescale-Prometheus
were stored under the values for the connector. This created
confusion for users because the prometheus chart has a separate
remoteWrite configuration option. This refactor aims to make
it clear that the configuration for the timescale-prometheus
remote_write will go into the config map override
of prometheus.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -8,7 +8,6 @@ timescaledb-single:
   enabled: true
   image:
     tag: pg12-ts1.7
-    pullPolicy: Always
   # create only a ClusterIP service
   loadBalancer: 
     enabled: false
@@ -75,7 +74,7 @@ prometheus:
       port: "9201"
       write:
         enabled: true 
-        # complete url = {protocol}://{host}:{port}/endpoint
+        # complete url = {protocol}://{host}:{port}/{endpoint}
         endpoint: "write"
         # write queue config 
         # see: https://prometheus.io/docs/practices/remote_write/
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -8,6 +8,7 @@ timescaledb-single:
   enabled: true
   image:
     tag: pg12-ts1.7
+    pullPolicy: Always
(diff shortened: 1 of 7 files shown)


chart/values.yaml
```diff
@@ -77,7 +69,28 @@ prometheus:
     # we provide a config map which will have the
     # timescale-prometheus connector already configured as
     # a remote endpoint for read/write (if enabled)
+    # this overrides the config map the prometheus chart 
+    # creates
     configMapOverrideName: "prometheus-config"
+    # values to be used in the override config map
+    # for a Timescale-Prometheus remote_write and remote_read config
+    timescaleRemote:
```
sgichohi, line 77: This name threw me off a bit; Given that we're simply trying to forward values to the prometheus chart, could we use their labels and just override those implicitly? I might be missing something:

``
prometheus:
  enabled: true
  alertmanager:
    enabled: false
  pushgateway:
    enabled: false
  server:
     remoteWrite: 
   ### prometheus things?
``

blagojts: It's not impossible, just that we'd have some more templating logic.
1. Current implementation for `server.remoteWrite` (if we keep timescaleRemote separate)
  - server.remoteWrite | toYaml | indent
2. If we were to make the url templated
  - writeConfig := range server.remoteWrite
     - key, value := range  $writeConfig
     - if $key == url then $key: {{ tpl $value }}
     - else $key: ($value | toYaml)
and I don't know if I can manage to align the properties for $key: ($value | toYaml)
because they can sometimes be simple values, but also maps

and then repeat it for remote_read

PR #24, opened 2020-05-06
Add static deployment and generation script

antekresic:
Users which want to use static deployment resource to deploy to a
Kubernetes cluster can use the generated script instead of needing to
depend on Helm services.

### DIFF
--- a/generate-deploy-script.sh
+++ b/generate-deploy-script.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+if [ -n "$DEBUG" ]; then
+	set -x
+fi
+
+# Default values.yaml file to provide to helm template command.
+FILE_ARG=./chart/values.yaml
+
+# Check if values file filepath was supplied.
+if [ "$#" -eq  "1" ]; then
+	if [ -f "$1" ]; then
+		FILE_ARG=$1
+	fi
+fi
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+DIR=$(cd $(dirname "${BASH_SOURCE}") && pwd -P)
+
+RELEASE_NAME=ts-observability
+NAMESPACE=ts-observability
+
+NAMESPACE_VAR="
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: $NAMESPACE
+  labels:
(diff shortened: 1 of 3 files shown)


generate-deploy-script.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+DIR=$(cd $(dirname "${BASH_SOURCE}") && pwd -P)
+
+RELEASE_NAME=timescale-observability
+NAMESPACE=timescale-observability
```
blagojts, line 14: With timescale-observability as the release name, some of the resources may end up having names > 63 characters and k8s will not deploy them
With all the resources already in the timescale-obs namespace we may opt for a smaller release name

antekresic: Ah didn't know that... So we should be using `timescale-obs`? I guess `ts-obs` is to short :grinning:

ismailyenigul:
It might be much better to provide  values.yml parameter to  ```helm template``` another customization in addition to the namespace.
Example values.yml to use external timescale database 

```
timescaledb-single:
    enabled: false
timescale-prometheus:
    enabled: true
    connection:
      user: postgres
      password: pasword
      host: remote-timescaledb.domain.com
      port: 5432
    service:
    loadBalancer:
      enabled: false
```

generate-deploy-script.sh
ismailyenigul: This is the example case for this error. Just tried with `0.1.0-alpha.3`  release
```
$ helm install timescale-observability timescale/timescale-observability --devel

Error: CronJob.batch "timescale-observability-timescale-prometheus-drop-chunk" is 
invalid: metadata.name: Invalid value: "timescale-observability-timescale-prometheus-drop-chunk": 
must be no more than 52 characters
```

cevian: let's do `ts-observability`?

cevian: but we should also fix this issue

blagojts: https://github.com/helm/helm/issues/6006 it seems helm is the source of the error, not k8s, a old check for older k8s versions. But they haven't removed it

PR #34, opened 2020-05-09
grafana and prometheus version update

ismailyenigul:


cevian:
Thanks @ismailyenigul

PR #42, opened 2020-06-17
Add CLI

archen2019:


ts-obs/cmd/timescaledbConnect.go
```diff
@@ -0,0 +1,133 @@
...
+        if err != nil {
+            return err
+        }
+
+        err = kubeExecCmd(masterpod, "", "psql -U postgres", os.Stdin, true)
+        if err != nil {
+            return err
+        }
+   } else {
+        pod := getPodObject(user, password)
```
cevian, line 68: flagging for discussion about whether we want to use a pod or local psql.

ts-obs/cmd/timescaledbConnect.go
```diff
@@ -0,0 +1,133 @@
...
+var timescaledbConnectCmd = &cobra.Command{
+	Use:   "connect",
+	Short: "Connects to the TimescaleDB/PostgreSQL database",
+	RunE:  timescaledbConnect,
+}
+
+func init() {
+	timescaledbCmd.AddCommand(timescaledbConnectCmd)
+    timescaledbConnectCmd.Flags().StringP("user", "u", "postgres", "user to login with")
+    timescaledbConnectCmd.Flags().StringP("password", "p", "", "environment variable where password is stored")
```
cevian, line 24: do we also want to support secrets here?

ts-obs/cmd/timescaledbGetPassword.go
```diff
@@ -0,0 +1,44 @@
...
+// timescaledbGetPasswordCmd represents the timescaledb get-password command
+var timescaledbGetPasswordCmd = &cobra.Command{
+	Use:   "get-password",
+	Short: "Gets the TimescaleDB/PostgreSQL password for a specific user",
+	RunE:  timescaledbGetPassword,
+}
+
+func init() {
+	timescaledbCmd.AddCommand(timescaledbGetPasswordCmd)
+    timescaledbGetPasswordCmd.Flags().StringP("user", "u", "postgres", "user whose password to get")
```
cevian, line 19: do we have more than 1 user in the secret now? Should we? @blagojts maybe you can weigh in?

ts-obs/tests/grafana_test.go
```diff
@@ -0,0 +1,71 @@
...
+    "testing"
+    "time"
+)
+
+func testGrafanaPortForward(t testing.TB, port string) {
+    var portforward *exec.Cmd
+
+    if port == "" {
+        fmt.Println("Running 'ts-obs grafana port-forward'")
+        portforward = exec.Command("/home/archen/go/bin/ts-obs", "grafana", "port-forward")
```
cevian, line 16: hard code home directory won't work.

ts-obs/tests/grafana_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func testGrafanaChangePass(t testing.TB, newpass string) {
+    var changepass *exec.Cmd
+
+    fmt.Printf("Running 'ts-obs grafana change-password %v'\n", newpass)
+    changepass = exec.Command("/home/archen/go/bin/ts-obs", "grafana", "change-password", newpass)
+
+    out, err := changepass.CombinedOutput()
+    if err != nil {
+        fmt.Println(string(out))
```
cevian, line 51:
```suggestion

```

ts-obs/tests/grafana_test.go
```diff
@@ -0,0 +1,71 @@
...
+        t.Fatal(err)
+    }
```
cevian, line 53: we should try to see if we can use the new password in the test, if possible.

ts-obs/tests/grafana_test.go
```diff
@@ -0,0 +1,71 @@
...
+    } else {
+        fmt.Printf("Running 'ts-obs grafana port-forward -p %v'\n", port)
+        portforward = exec.Command("/home/archen/go/bin/ts-obs", "grafana", "port-forward", "-p", port)
+    }
+
+    err := portforward.Start()
+    if err != nil {
+        t.Fatal(err)
+    }
+    time.Sleep(4 * time.Second)
```
cevian, line 26: can we test the port is open or something?

ts-obs/tests/installation_test.go
```diff
@@ -0,0 +1,145 @@
...
+    "os/exec"
+    "testing"
+    "time"
+)
+
+func testInstall(t testing.TB, name string, filename string) {
+    var install *exec.Cmd
+
+    if name == ""  && filename == "" {
+        fmt.Println("Running 'ts-obs install'")
```
cevian, line 14: all these prints will eventually have to be deleted I think.

ts-obs/tests/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+        deleteKube()
+        done <- true
+        os.Exit(1)
+    }()
+
+    startKube()
+    installObs()
+
+    fmt.Println("Sleep to wait for pods to initialize")
+    time.Sleep(300 * time.Second)
```
cevian, line 82: let's make sure the pods are indeed up.

ts-obs/tests/installation_test.go
```diff
@@ -0,0 +1,145 @@
...
+    testHelmUninstall(t, "f1")
+    testHelmInstall(t, "f2", "./testdata/f2.yml")
+    testUninstall(t, "f2")
+    testHelmInstall(t, "f3", "./testdata/f3.yml")
+    testHelmUninstall(t, "f3")
+    testInstall(t, "f4", "./testdata/f4.yml")
+    testUninstall(t, "f4")
+
+    testInstall(t, "", "")
+    fmt.Println("Sleep to wait for pods to initialize")
```
cevian, line 143: why do we have this here? Seems like TestMain() already does this

ts-obs/tests/installation_test.go
```diff
@@ -0,0 +1,145 @@
...
+    } else {
+        fmt.Printf("Running 'ts-obs helm uninstall -n %v'\n", name)
+        uninstall = exec.Command("/home/archen/go/bin/ts-obs", "uninstall", "-n", name)
+    }
+
+    out, err := uninstall.CombinedOutput()
+    if err != nil {
+        fmt.Println(string(out))
+        t.Fatal(err)
+    }
```
cevian, line 91: verification of some sort?

ts-obs/tests/portforward_test.go
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+    }
+
+    err := portforward.Start()
+    if err != nil {
+        t.Fatal(err)
+    }
+    time.Sleep(4 * time.Second)
+    portforward.Process.Signal(syscall.SIGINT)
```
cevian, line 57: some verification would be nice here...

ts-obs/cmd/timescaledbGetPassword.go
archen2019: We currently have a postgres user and an admin user in the secret.

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,83 @@
...
+// helmInstallCmd represents the helm install command
+var helmInstallCmd = &cobra.Command{
+	Use:   "install",
+	Short: "Installs Timescale Observability",
+	RunE: helmInstall,
+}
+
+func init() {
+	helmCmd.AddCommand(helmInstallCmd)
+    helmInstallCmd.Flags().StringP("name", "n", "ts-obs", "Release name")
```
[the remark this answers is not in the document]
archen2019: If we want the option for people to install two or more Helm charts, then release name can't be set to default, or there's going to be a conflict. Or we can maybe automatically check if there is already an instance of the chart installed. The global flag solution seems like it would be annoying to type every time. ENV variables could work but then they would need to be manually changed every time you want to do operations in a different set of pods.

ts-obs/cmd/chunkIntervalSet.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func chunkIntervalSet(cmd *cobra.Command, args []string) error {
+    var err error
+
+    if len(args) != 2 {
+        return errors.New("\"ts-obs metrics set-metric-chunk\" requires 2 arguments")
+    }
+
+    if os.Getenv("PGPASSWORD_POSTGRES")  == "" {
```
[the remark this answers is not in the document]
archen2019: The secret doesn't update if the password is changed.

ts-obs/cmd/kubectl.go
```diff
@@ -0,0 +1,302 @@
...
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+    "k8s.io/apimachinery/pkg/labels"
+)
+
+func kubeInit() (kubernetes.Interface, *rest.Config) {
+    var err error
+
+    config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: os.Getenv("HOME") + "/.kube/config"},
+		&clientcmd.ConfigOverrides{CurrentContext: "minikube"}).ClientConfig()
```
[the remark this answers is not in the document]
archen2019: Unnecessary config, removed.

ts-obs/cmd/portForward.go
```diff
@@ -0,0 +1,74 @@
...
+    if err != nil {
+        return err
+    }
+
+    err = kubePortForwardService(serviceName, prometheus, 9090)
+    if err != nil {
+        return err
+    }
+
+    select {}
```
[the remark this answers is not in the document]
archen2019: Infinite loop so that port forwarding continues to happen.

ts-obs/cmd/portForward.go
```diff
@@ -0,0 +1,74 @@
...
+
```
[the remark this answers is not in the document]
archen2019: Yeah, infinite loops so that the port forwarding doesn't stop right away.

ts-obs/tests/installation_test.go
archen2019: It's testing the install/uninstall commands, so after the final install in this test, we need to wait for the pods to initialize again, in case there are more tests afterward.

ts-obs/cmd/root.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.ts-obs.yaml)")
+
+	// Cobra also supports local flags, which will only run
+	// when this action is called directly.
+	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
```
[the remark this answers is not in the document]
archen2019: Was automatically generated by cobra, removing it shouldn't affect anything.

ts-obs/cmd/timescaledbConnect.go
```diff
@@ -0,0 +1,133 @@
...
+        return err
+    }
+
+    var master bool
+    master, err = cmd.Flags().GetBool("master")
+    if err != nil {
+        return err
+    }
+
+    if password == "" == !master {
```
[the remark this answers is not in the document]
archen2019: We need exactly one of them to be true, since connecting through master automatically logs you in as postgres. Maybe change to an XOR to make that clearer.

ts-obs/README.md
```diff
@@ -0,0 +1,69 @@
+# Timescale Observability CLI
+
+This is a CLI tool for installing and managing the Timescale Observability Helm chart. 
+
+## Quick Start
+
+__Dependencies__: [Helm](https://helm.sh/docs/intro/install/)
+
+To install the CLI, run `go install` from inside the `ts-obs` folder.
```
blagojts, line 9: go not listed as dependency

ts-obs/cmd/chunkInterval.go
```diff
@@ -0,0 +1,15 @@
+package cmd
```
blagojts, line 1: go file convention favors lowercase names with underscores `chunk_interval`

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,56 @@
+package cmd
```
blagojts, line 1: file not formatted

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,56 @@
...
+func init() {
+	chunkIntervalCmd.AddCommand(chunkIntervalResetCmd)
+}
+
+func chunkIntervalReset(cmd *cobra.Command, args []string) error {
+    var err error
+
+    if len(args) != 1 {
+        return errors.New("\"ts-obs metrics reset-metric-chunk-interval\" requires 1 arguments")
+    }
```
blagojts, line 29: Cobra has a way of checking arguments for you, check out (https://github.com/spf13/cobra#positional-and-custom-arguments, and the example bellow it with `cobra.MinimumNArgs(1)`

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,56 @@
...
+
+    if os.Getenv("PGPASSWORD_POSTGRES")  == "" {
+        return errors.New("Password for postgres user must be set in environment variable PGPASSWORD_POSTGRES")
+    }
```
blagojts, line 33: Postgres environment variables are more or less known/codified maybe we should use the same (PGPASSWORD) (https://www.postgresql.org/docs/current/libpq-envars.html)

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,56 @@
...
+
+    metric := args[0]
+
+    err = kubePortForwardPod("ts-obs-timescaledb-0", 5432, 5432)
+    if err != nil {
+        return err
+    }
```
blagojts, line 40: We should wrap this error into something like "could not reset chunk interval: <internal error here>"

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,56 @@
...
+
+    var pool *pgxpool.Pool
+    pool, err = pgxpool.Connect(context.Background(), "postgres://postgres:" + os.Getenv("PGPASSWORD_POSTGRES") + "@localhost:5432/postgres")
```
blagojts, line 43: user and database should be configurable (forge doesn't use postgres and we want to encourage users to use obs with forge)

ts-obs/cmd/chunkIntervalSet.go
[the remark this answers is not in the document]
blagojts: people should update their secrets, because when spawning more pods the secret will be used to populate the environment variables, changing the pass in the db only will break you're whole obs stack

ts-obs/cmd/chunkIntervalSet.go
```diff
@@ -0,0 +1,67 @@
...
+    metric := args[0]
+    var chunk_interval time.Duration
+    chunk_interval, err = time.ParseDuration(args[1])
+    if err != nil {
+        return err
+    }
+
+    if chunk_interval.Microseconds() < 1 {
+        return errors.New("Chunk interval must be at least 1 microsecond")
+    }
```
blagojts, line 46: I know it's possible, but maybe we should limit people from actually setting chunks that last less than a minute?

ts-obs/cmd/helmGetYaml.go
```diff
@@ -0,0 +1,40 @@
...
+    "fmt"
+    "os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+// helmGetYamlCmd represents the helm get-yaml command
+var helmGetYamlCmd = &cobra.Command{
+	Use:   "get-yaml",
+	Short: "Writes the current configuration YAML to console",
```
blagojts, line 14:
```suggestion
	Short: "Prints the default timescale-obserability values to console",
```

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,83 @@
...
+        return err
+    }
+
+    w := io.Writer(os.Stdout)
+
+    addchart := exec.Command("helm", "repo", "add", "timescale", "https://charts.timescale.com")
+
+    addchart.Stdout = w
+    addchart.Stderr = w
+    fmt.Println("Adding Timescale Helm chart")
```
blagojts, line 51:
```suggestion
    fmt.Println("Adding Timescale Helm Repository")
```

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,83 @@
...
+    err = addchart.Run()
+    if err != nil {
+        return err
+    }
+
+    update := exec.Command("helm", "repo", "update")
+
+    update.Stdout = w
+    update.Stderr = w
+    fmt.Println("Updating local chart info")
```
blagojts, line 61:
```suggestion
    fmt.Println("Fetching updates from repository")
```

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    install.Stdout = w
+    install.Stderr = w
+    fmt.Println("Installing Timescale Observability")
+    err = install.Run()
+    if err != nil {
+        return err
+    }
+
```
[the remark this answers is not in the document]
blagojts: ^ That would actually help a lot of our users, it will prevent them from expecting everything to be working while the db is still booting up

ts-obs/tests/installation_test.go
blagojts: I like them :\

ts-obs/cmd/chunkIntervalSet.go
archen2019: Sounds reasonable, I can't think of any use case where anyone would need to go below a minute.

ts-obs/cmd/helmInstall.go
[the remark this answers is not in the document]
archen2019: Done.

ts-obs/cmd/timescaledbConnect.go
archen2019: Making a pod takes a while sometimes, portforwarding for local psql is much faster.

ts-obs/tests/installation_test.go
archen2019: I think the quit flag is a good idea.

ts-obs/tests/grafana_test.go
archen2019: This only prints when there's an error, is there a need to remove?

ts-obs/cmd/chunkIntervalReset.go
archen2019: I was following the naming convention that we had in our notes when the Helm install happens, since we have an admin and a postgres user. May need to discuss whether the admin user is necessary.

ts-obs/tests/installation_test.go
archen2019: Using the default go test -v flag instead.

ts-obs/cmd/chunkInterval.go
archen2019: Cobra specifically asks for command names in camel case, which is why I left it that way.

ts-obs/cmd/kubectl.go
```diff
@@ -1,302 +1,332 @@
...
+	var err error
 
-    config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{ExplicitPath: os.Getenv("HOME") + "/.kube/config"},
-		&clientcmd.ConfigOverrides{CurrentContext: "minikube"}).ClientConfig()
-    if err != nil {
-        log.Fatal(err)
-    }
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: os.ExpandEnv("$HOME/.kube/config")},
```
blagojts, line 28: maybe we should put $HOME in a var so we can override it when building for Windows? (we used this for timescale-prometheus https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications)

ts-obs/cmd/kubectl.go
```diff
@@ -1,302 +1,332 @@
...
+	client, _ := KubeInit()
+
+	labelSelector := metav1.LabelSelector{MatchLabels: labelmap}
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
+	}
+
+	pvcs, err := client.CoreV1().PersistentVolumeClaims(namespace).List(context.Background(), listOptions)
+	if err != nil {
+		return []string{}, err
```
blagojts, line 103:
```suggestion
		return nil, err
```

ts-obs/cmd/timescaledbConnect.go
```diff
@@ -105,26 +112,26 @@ func getPodObject(user, password string) *corev1.Pod {
...
+					},
+					Args: []string{
 						"-U",
-                        user,
-                        "-h",
-                        "ts-obs.default.svc.cluster.local",
-                        "postgres",
+						user,
+						"-h",
+						name + ".default.svc.cluster.local",
```
blagojts, line 133: "default" is the namespace

ts-obs/cmd/timescaledbGetPassword.go
```diff
@@ -11,34 +10,43 @@ import (
...
+		return err
+	}
+
+	secret, err := KubeGetSecret(namespace, name+"-timescaledb-passwords")
+	if err != nil {
+		return err
+	}
+
+	pass := secret.Data[user]
+	fmt.Printf("Password: %v\n", string(pass))
```
blagojts, line 49: The grafana get pass cmd just prints out the password, we should unify the prints (either "Password: xxxx" or just "xxxx") but I'm not sure which is preferred by users :\

archen2019: I think it makes more sense to just print the password, it makes it easier to copy paste

ts-obs/cmd/grafanaGetPassword.go
```diff
@@ -0,0 +1,45 @@
...
+	var err error
+
+	var name string
+	name, err = cmd.Flags().GetString("name")
+	if err != nil {
+		return err
+	}
+
+	var namespace string
+	namespace, err = cmd.Flags().GetString("namespace")
```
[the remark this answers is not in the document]
archen2019: Yeah, it'a a persistent flag set in the root, but it's read for each command.

ts-obs/cmd/helmUninstall.go
```diff
@@ -15,75 +15,89 @@ import (
...
+	pvcnames, err := KubeGetPVCNames(namespace, map[string]string{"release": name})
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Removing Persistent Volume Claims")
+	for _, s := range pvcnames {
+		err = KubeDeletePVC(namespace, s)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
archen2019: There's only two of them, maybe not super necessary.

ts-obs/cmd/chunkIntervalSet.go
```diff
@@ -1,67 +1,95 @@
...
+	err = KubePortForwardPod(namespace, podName, LISTEN_PORT_TSDB, FORWARD_PORT_TSDB)
+	if err != nil {
+		return err
+	}
+
+	var pool *pgxpool.Pool
+	pool, err = pgxpool.Connect(context.Background(), "postgres://"+user+":"+os.Getenv("PGPASSWORD")+"@localhost:"+strconv.Itoa(LISTEN_PORT_TSDB)+"/"+dbname)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
archen2019: Either way the code will be essentially 1-2 lines, and with a function I would have to pass a lot of arguments. Not sure if its worth it, even though the code is repeated.

ts-obs/cmd/chunkIntervalReset.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	pass := string(secret.Data[user])
+
+	podName, err := KubeGetPodName(namespace, map[string]string{"release": name, "role": "master"})
+	if err != nil {
+		return fmt.Errorf("could not reset chunk interval for %v: %w", metric, err)
+	}
+
+	err = KubePortForwardPod(namespace, podName, LISTEN_PORT_TSDB, FORWARD_PORT_TSDB)
```
cevian, line 67: for these types of "internal" port forward the local port should be 0. and you should use https://godoc.org/k8s.io/client-go/tools/portforward#PortForwarder.GetPorts to get the local port actually used. The reason is LISTEN_PORT_TSDB may already be in use locally by something else, and so it's better to have the OS just pick a free port.

ts-obs/cmd/chunkIntervalSetDefault.go
```diff
@@ -0,0 +1,96 @@
...
+
+	var pool *pgxpool.Pool
+	pool, err = pgxpool.Connect(context.Background(), "postgres://"+user+":"+pass+"@localhost:"+strconv.Itoa(LISTEN_PORT_TSDB)+"/"+dbname)
+	if err != nil {
+		return fmt.Errorf("could not set default chunk interval: %w", err)
+	}
+	defer pool.Close()
+
+	fmt.Printf("Setting default chunk interval to %v\n", chunk_interval)
+	_, err = pool.Exec(context.Background(), "SELECT prom_api.set_default_chunk_interval(INTERVAL '1 second' * $1)", strconv.FormatFloat(chunk_interval.Seconds(), 'f', -1, 64))
```
cevian, line 90:
```suggestion
	_, err = pool.Exec(context.Background(), "SELECT prom_api.set_default_chunk_interval($1::INTERVAL)", chunk_interval)
```

cevian: Other places too

ts-obs/cmd/grafanaChangePassword.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	grafanaPod, err := KubeGetPodName(namespace, map[string]string{"app.kubernetes.io/instance": name, "app.kubernetes.io/name": "grafana"})
+	if err != nil {
+		return fmt.Errorf("could not change Grafana password: %w", err)
+	}
+
+	err = KubeExecCmd(namespace, grafanaPod, "grafana", "grafana-cli admin reset-admin-password "+password, nil, false)
+	if err != nil {
+		return fmt.Errorf("could not change Grafana password: %w", err)
```
cevian, line 56: let's at least attempt to update the secret back to the original value if there is a failure here.

ts-obs/cmd/kubectl.go
```diff
@@ -0,0 +1,339 @@
+package cmd
+
```
cevian, line 2: For a whole bunch of these functions I wonder if the code would be cleaner if instead fo getting the namespace and name as parameters, it would be better to do the following inside these functions.

Would that remove a lot of code duplication?

 ```	var name string
	name, err = cmd.Flags().GetString("name")
	if err != nil {
		return fmt.Errorf("could not find release name: %w", err)
	}

	var namespace string
	namespace, err = cmd.Flags().GetString("namespace")
	if err != nil {
		return fmt.Errorf("could not get namespace: %w", err)
	}```

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return fmt.Errorf("could not install Timescale Observability: %w", err)
+	}
+
+	var install *exec.Cmd
+	if DEVEL {
+		if file == "" {
+			install = exec.Command("helm", "install", name, "timescale/timescale-observability", "--devel")
+		} else {
+			install = exec.Command("helm", "upgrade", "--install", name, "--values", file, "timescale/timescale-observability", "--devel")
```
cevian: do we want ts-obs install to upgrade if the release exists. I think not. I'd rather add a ts-obs upgrade command that installs if things don't exist than the other way around.

ts-obs/cmd/kubectl.go
```diff
@@ -0,0 +1,339 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func KubeWaitOnPod(namespace string, podName string) error {
+	client, _ := KubeInit()
+
+	for {
```
cevian, line 271: we should have some limit here. Maybe 10 min?

ts-obs/cmd/kubectl.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func KubeWaitOnPod(namespace string, podName string) error {
+	client, _ := KubeInit()
+
```
cevian, line 270: We should print "Waiting for podname" if this is called during install

ts-obs/cmd/kubectl.go
```diff
@@ -0,0 +1,339 @@
...
+	for {
+		pod, err := client.CoreV1().Pods(namespace).Get(context.Background(), podName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		if pod.Status.Phase != corev1.PodPending && pod.Status.Phase != corev1.PodFailed {
```
cevian, line 276: We should print "podname has started" if this is called during install

ts-obs/cmd/kubectl.go
cevian: This gives people more visibility as to what's going on

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+
+	for _, pod := range pods {
+		err = KubeWaitOnPod(namespace, pod.Name)
+		if err != nil {
+			return err
+		}
+	}
+
```
cevian, line 125: Let's print something about "Everything started successfully" or something

ts-obs/cmd/helmUninstall.go
```diff
@@ -0,0 +1,105 @@
...
+	err = KubeDeleteEndpoint(namespace, name)
+	if err != nil {
+		fmt.Println(err, ", skipping")
+	}
+
+	if !pvc {
+		return nil
+	}
+
+	fmt.Println("Getting Persistent Volume Claims")
```
cevian, line 95: Deleting a PVC can delete the data associated with it. So ts-obs uninstall should have a --remove-data flag that is false by default. The pvc deletion happens if the flag is true, if its false, we print a message saying that the data was retained and how to delete it. As for how to delete it: Maybe running ts-obs a second time with the flag should do it or maybe we should have a separate sub-command. What do you think?

ts-obs/cmd/portForward.go
```diff
@@ -0,0 +1,100 @@
...
+	serviceName, err = KubeGetServiceName(namespace, map[string]string{"release": name, "app": "prometheus", "component": "server"})
+	if err != nil {
+		return fmt.Errorf("could not port-forward: %w", err)
+	}
+
+	err = KubePortForwardService(namespace, serviceName, prometheus, FORWARD_PORT_PROM)
+	if err != nil {
+		return fmt.Errorf("could not port-forward: %w", err)
+	}
+
```
cevian, line 96: Will an informational message be printed with which ports are being forwarded?

ts-obs/tests/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+		os.Exit(1)
+	}
+}
+
+func installObs() {
+	var err error
+
+	log.Println("Installing Timescale Observability")
+
+	obsinstall := exec.Command("ts-obs", "install")
```
cevian, line 29: I think the tests will more likely find bugs if our main install uses a non-default namespace and non-default release name

ts-obs/tests/main_test.go
```diff
@@ -0,0 +1,89 @@
...
+		stopKube()
+		deleteKube()
+		done <- true
+		os.Exit(1)
+	}()
+
+	startKube()
+	installObs()
+
+	os.Setenv("PGPASSWORD_POSTGRES", "tea")
```
cevian, line 80: do we need this anymore?

ts-obs/tests/metrics_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	os.Setenv("PGPASSWORD_POSTGRES", "tea")
+
+	pool, err := pgxpool.Connect(context.Background(), "postgres://postgres:"+os.Getenv("PGPASSWORD_POSTGRES")+"@localhost:5433/postgres")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = pool.QueryRow(context.Background(),
+		`SELECT EXTRACT(epoch FROM _prom_catalog.get_metric_retention_period($1))`,
```
cevian, line 112: I wonder if we should have a cli command to get the retention period for a metric

ts-obs/tests/metrics_test.go
```diff
@@ -0,0 +1,212 @@
...
+	time.Sleep(10 * time.Second)
+
+	os.Setenv("PGPASSWORD_POSTGRES", "tea")
+
+	pool, err := pgxpool.Connect(context.Background(), "postgres://postgres:"+os.Getenv("PGPASSWORD_POSTGRES")+"@localhost:5433/postgres")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = pool.QueryRow(context.Background(),
```
cevian, line 143: maybe add a cli command to get the chunk interval for a metric?

ts-obs/cmd/chunkIntervalSet.go
```diff
@@ -0,0 +1,97 @@
...
+
+	pass := string(secret.Data[user])
+
+	podName, err := KubeGetPodName(namespace, map[string]string{"release": name, "role": "master"})
+	if err != nil {
+		return fmt.Errorf("could not set chunk interval for %v: %w", metric, err)
+	}
+
+	err = KubePortForwardPod(namespace, podName, LISTEN_PORT_TSDB, FORWARD_PORT_TSDB)
```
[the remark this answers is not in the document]
archen2019: The code from 33 to 64 is reading flags and error catching. I think from 65 to here can be rewritten as a separate function which can then be called.

ts-obs/cmd/portForward.go
archen2019: The information shows up but doesn't specify which service/pod it corresponds to. Just added a clearer message.

ts-obs/tests/main_test.go
archen2019: no

ts-obs/cmd/grafanaChangePassword.go
```diff
@@ -0,0 +1,53 @@
...
+	fmt.Println("Changing password...")
+	grafanaPod, err := KubeGetPodName(namespace, map[string]string{"app.kubernetes.io/instance": name, "app.kubernetes.io/name": "grafana"})
+	if err != nil {
+		return fmt.Errorf("could not change Grafana password: %w", err)
+	}
+
+	err = KubeExecCmd(namespace, grafanaPod, "grafana", "grafana-cli admin reset-admin-password "+password, nil, false)
+	if err != nil {
+		secret.Data["admin-password"] = oldpassword
+		KubeUpdateSecret(namespace, secret)
```
cevian, line 48:
```suggestion
		_ = KubeUpdateSecret(namespace, secret)
```
To signify that the error is intentionally ignored

ts-obs/cmd/helmInstall.go
```diff
@@ -0,0 +1,115 @@
...
+	update.Stdout = w
+	update.Stderr = w
+	fmt.Println("Fetching updates from repository")
+	err = update.Run()
+	if err != nil {
+		return fmt.Errorf("could not install Timescale Observability: %w", err)
+	}
+
+	var install *exec.Cmd
+	if DEVEL {
```
cevian, line 61: This code would be cleaner as something like 
```
cmds := []string{"helm", "install"}
if namespace != "default" {
   cmds = append(cmds, "-n", namespace)
}
if file != "" {
   cmds = append(cmd, "--values", file)
}
...
install = exec.Command(cmds...)
```

ts-obs/cmd/helmUninstall.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		fmt.Println(err, ", skipping")
+	}
+
+	err = KubeDeleteEndpoint(namespace, name)
+	if err != nil {
+		fmt.Println(err, ", skipping")
+	}
+
+	fmt.Println("Data still remains. To delete data as well, run 'ts-obs helm delete-data'")
```
cevian, line 72: I would provide an optional flag (`--delete-data`) to this command (default false) that if true, does this automatically

ts-obs/cmd/root.go
```diff
@@ -0,0 +1,81 @@
...
+compressed long-term store for time series metrics from Prometheus. This
+application is a CLI tool that allows users to quickly access the different
+components of Observability.`,
+	SilenceUsage: true,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+
+		name, err = cmd.Flags().GetString("name")
+		if err != nil {
+			return fmt.Errorf("could not change Grafana password: %w", err)
```
cevian, line 30: incorrect error msg

ts-obs/cmd/root.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		namespace, err = cmd.Flags().GetString("namespace")
+		if err != nil {
+			return fmt.Errorf("could not change Grafana password: %w", err)
```
cevian, line 35: incorrect error message

ts-obs/cmd/retentionGet.go
```diff
@@ -0,0 +1,59 @@
...
+
+	pool, err := OpenConnectionToDB(namespace, name, user, dbname, FORWARD_PORT_TSDB)
+	if err != nil {
+		return fmt.Errorf("could not get retention period for %v: %w", metric, err)
+	}
+	defer pool.Close()
+
+	fmt.Printf("Getting retention period for %v\n", metric)
+	var secs int
+	err = pool.QueryRow(context.Background(), "SELECT EXTRACT(epoch FROM _prom_catalog.get_metric_retention_period($1))", metric).Scan(&secs)
```
blagojts, line 50: `get_metric_retention_period` returns an pg interval type, pgx should be able to read this value in a time.Duration or you can read it in `pgtype.Interval` (package in pgx) and then assign it to a time.Duration type

ts-obs/tests/main_test.go
```diff
@@ -6,8 +6,12 @@ import (
 	"os/exec"
 	"os/signal"
 	"testing"
+	"time"
 )
 
+const RELEASE_NAME = "gg"
+const NAMESPACE = "ns"
```
blagojts, line 13: do we have tests that do concurrent installs (with diff release names) or it will require too much resources from the minikube vm?

ts-obs/cmd/kubectl.go
```diff
@@ -98,30 +110,56 @@ func KubeGetPVCNames(namespace string, labelmap map[string]string) ([]string, er
...
+	if err != nil {
+		return nil, err
 	}
 
-	return names, nil
+	return secret, nil
 }
 
-func KubeGetAllPods(namespace string) ([]corev1.Pod, error) {
+func KubeGetAllPods(namespace string, name string) ([]corev1.Pod, error) {
```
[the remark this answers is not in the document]
archen2019: The majority of our pods have the label "release: name", it would definitely make it easier if all the pods had it.

ts-obs/tests/main_test.go
archen2019: I think its possible, but would require a good amount of refactoring. Maybe for the next pr?

ts-obs/cmd/retentionGet.go
archen2019: This doesn't work since the pg interval has days/months which can't be scanned into a time.Duration. I just changed it to extract the number of days and print that, since we set retention in days.

PR #45, opened 2020-07-16
Make CLI compatible with Forge

archen2019:
- Add port number to GF_DATABASE_HOST
- Read dbname and port number from values file
- OpenConnectionToDB now detects and supports external DB

### DIFF
--- a/chart/templates/grafana-datasources-sec.yaml
+++ b/chart/templates/grafana-datasources-sec.yaml
@@ -30,9 +30,8 @@ stringData:
 {{- end -}}
 {{ if $tsEnabled -}}
 {{- $isDefault := not $promEnabled -}}
-{{- $root := . -}}
-{{- $host := tpl .Values.grafana.timescale.database.host $root -}}
-{{- $port := index .Values "timescale-prometheus" "connection" "port" | int -}}
+{{- $host := tpl .Values.grafana.timescale.database.host $ -}}
+{{- $port := .Values.grafana.timescale.database.port | int -}}
 {{ with .Values.grafana.timescale.datasource }}
       - name: TimescaleDB
         url: {{ printf "%s:%d" $host $port }}
--- a/chart/templates/grafana-datasources-sec.yaml
+++ b/chart/templates/grafana-datasources-sec.yaml
@@ -31,9 +31,11 @@ stringData:
 {{ if $tsEnabled -}}
 {{- $isDefault := not $promEnabled -}}
 {{- $root := . -}}
+{{- $host := tpl .Values.grafana.timescale.database.host $root -}}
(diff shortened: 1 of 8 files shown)


ts-obs/cmd/pgx.go
```diff
@@ -29,22 +49,29 @@ func OpenConnectionToDB(namespace, name, user, dbname string, remote int) (*pgxp
 		return nil, errors.New("user not found")
 	}
 
-	podName, err := KubeGetPodName(namespace, map[string]string{"release": name, "role": "master"})
-	if err != nil {
-		return nil, err
-	}
+	if strings.HasSuffix(host, ".svc.cluster.local") {
```
cevian, line 52: is this the best way to determine this? Can we instead see if the db pod exists and port-forward if it does (not port-forwarding if it doesn't)? @blagojts do you have any ideas here?

chart/templates/grafana-db-user-conf.yaml
```diff
@@ -21,6 +21,7 @@ data:
         RAISE NOTICE 'role {{ $timescaleAsDB.user }} already exists, skipping create';
       END
     $$;
+    GRANT {{ $timescaleAsDB.user }} TO {{ .Values.grafana.timescale.adminUser }};
```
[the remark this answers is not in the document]
archen2019: The CREATE SCHEMA command after errors out otherwise. We used to use the superuser so it wasn't necessary.

ts-obs/cmd/pgx.go
archen2019: there are probably better ways to do this, but i can't think of any realistic instance where it would fail. checking the pod is possible, but i dont know if the extra complication is necessary

blagojts: For services in the same k8s cluster you can just specify the service name `ts-obs` or even `ts-obs.namespace` and it will still be resolved by the DNS service, the svc.cluster.local is not necessary. So this would fail in those cases

chart/templates/grafana-datasources-sec.yaml
```diff
@@ -31,9 +31,11 @@ stringData:
 {{ if $tsEnabled -}}
 {{- $isDefault := not $promEnabled -}}
 {{- $root := . -}}
+{{- $host := tpl .Values.grafana.timescale.database.host $root -}}
+{{- $port := index .Values "timescale-prometheus" "connection" "port" | int -}}
```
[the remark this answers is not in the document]
archen2019: Added, but we can't do it the same way as we did the hostname, since the port is an int. Helm conventions say not to put quotes around ints that arent environment variables, so we will have to just edit the port number in 3 places.

Issue #47, opened 2020-07-28
Please add --devel to the docs for every helm command.


seletz:
# Problem

The docs fail to mention `--devel` in the step where you'd download and customise your `values.yaml`.  

Ppl like me who copy-paste tutorial code will get confused, helm won't find a valid chart.

As mentioned by @mike in the slack channel the beta will come soon, but maybe it will save others some headaches.

PR coming.

PR #48, opened 2020-07-28
Updated docs to add --devel to helm commands

seletz:
The docs failed to mention `--devel` in one instance, this got me
confused as the error given by helm in that case is not really helpful.


seletz:
Should fix #47

PR #52, opened 2020-08-18
Increase default WAL and data disk size

### LINKED ISSUE #38: Increase default wal and data disk size
Need to experiment a bit but data should probably be ~50GB wal maybe 5GB

archen2019:
Fixes #38

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -14,6 +14,11 @@ timescaledb-single:
     enabled: false
   # number or TimescaleDB pods to spawn (default is 3, 1 for no HA)
   replicaCount: 1
+  persistentVolumes:
+    data:
+      size: 150Gi
+    wal:
+      size: 20Gi
 
 # Values for configuring the deployment of the Timescale Prometheus Connector
 # The charts README is at:


chart/values.yaml
```diff
@@ -14,6 +14,15 @@ timescaledb-single:
...
   replicaCount: 1
+  persistentVolumes:
+    data:
+      size: 50Gi
+  patroni:
+    bootstrap:
+      dcs:
+        postgresql:
+          parameters:
+            max_wal_size: 5120
```
cevian, line 25: I don't think this is the right setting. I think what we want to set is `persistentVolumes.wal.size` let's set that to 20Gi actually

chart/values.yaml
```diff
@@ -14,6 +14,15 @@ timescaledb-single:
     enabled: false
   # number or TimescaleDB pods to spawn (default is 3, 1 for no HA)
   replicaCount: 1
+  persistentVolumes:
+    data:
+      size: 50Gi
```
cevian, line 19: Based on what we see on forge lets make this 150Gi actually

PR #57, opened 2020-08-26
Change README to emphasize CLI tool.

cevian:
For reviewers: This reorganizes the readme so that the front-page readme is much shorter and to-the-point. It also emphasizes CLI usage. Details about the helm chart are moved into a secondary readme under chart/. Some other changes too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,219 +1,31 @@
 # tobs - The Observability Stack for Kubernetes
 
-A Helm chart for deploying Prometheus configured to use TimescaleDB as compressed long-term store for time-series metrics through the Timescale-Prometheus Connector.
+Tobs deploys an observability stack consisting of Prometheus, TimescaleDB, and Grafana to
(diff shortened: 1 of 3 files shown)


laceybutler:
**Stream of consciousness feedback**

* Per Slack, I suggest including a bit about the *problem* the tool solves in the intro and/or common applications (i.e., when and why someone would use this - is it easier to deploy? more comprehensive? more extensible?). 

* Additionally, I'd suggest making an "Installation with Helm (Recommended)" header for the first installation option and WHY you recommend this option v. the other (same size as the "Installation w/o Helm" section heading.

* Should y'all include anything about what this will *do* for folks and/or point to additional docs (e.g., after I run the command in my terminal, what happens?). 

`After setting up tobs, run the following to install the tobs helm charts into your kubernetes cluster`

^ Does it make sense to clarify installation as a two-step process: "set up" (what this does) and "install/deploy helm chart to your Kubernetes cluster(s)" (and what this does/what users should expect)?

PR #60, opened 2020-09-04
Prevent potential problems with partial content/download

cevian:
Put everything inside a function so that nothing actually done until full content is
downloaded.

See https://www.arp242.net/curl-to-sh.html


JLockerman (approved):
🎉

PR #64, opened 2020-09-23
Add Promlens to the stack

cevian:


chart/templates/promlens-depoloyment.yaml
```diff
@@ -0,0 +1,35 @@
...
+    metadata:
+      labels:
+        app: {{ template "tobs.fullname" . }}
+        component: promlens
+    spec:
+      containers:
+        - image: {{ .Values.promlens.image }}
+          imagePullPolicy: IfNotPresent
+          name: promlens
+          command: ["/promlens", "--web.default-prometheus-url", "{{ tpl .Values.promlens.defaultPrometheusUrl . }}" ]
```
Harkishen-Singh, line 27: Should we use `--log.level=<severity-var>`, `--log.format=logfmt` and `--web.external-url=8080` (assuming if in future this changes in case)?

chart/templates/promlens-depoloyment.yaml
```diff
@@ -0,0 +1,35 @@
...
+          {{- if .Values.promlens.resources }}
+          resources:
+            {{- toYaml .Values.promlens.resources | nindent 12 }}
+          {{- end }}
+          ports:
+            - containerPort: 8080
+              name: ui
```
Harkishen-Singh, line 34: Not sure, but is `ui` a good way? I mean, in terms of UI, we have grafana and promlens. How about `promlens-container-port`?

cli/cmd/promlensPortForward.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+const LISTEN_PORT_PROMLENS = 8081
+const FORWARD_PORT_PROMLENS = 8080
+const LISTEN_PORT_CONNECTOR = 9201
+const FORWARD_PORT_CONNECTOR = 9201
```
Harkishen-Singh, line 12: Isn't a single const enough? Reminds me about tslint standards.

Harkishen-Singh (changes requested):

cli/cmd/promlensPortForward.go
cevian: Logically they are different so I think better to keep separate

chart/templates/promlens-depoloyment.yaml
cevian: Changed to `promlens-web-interface`

chart/templates/promlens-depoloyment.yaml
cevian: added "--web.listen-address", ":8080" but I think others better kept default

chart/templates/promlens-depoloyment.yaml
Harkishen-Singh: that is perfect

Harkishen-Singh (approved):
AFAIK, lgtm!

PR #80, opened 2020-11-17
Make install-cli compatible with dash on Linux

### LINKED ISSUE #66: tobs CLI default installation fails
```
$ curl --proto '=https' --tlsv1.2 -sSLf  https://tsdb.co/install-tobs-sh | sh
sh: 31: Syntax error: redirection unexpected
```

Because of the redirection opertion at   `if grep -Fxq "${checksum}" <<< "${checksumlist}"; then`

Instead of using sh, using bash is making this work.

We should either update README.md to use `curl <> | bash` or find an alternative to redirection operation.

I am using ubuntu 20.04.

blagojts:
The <<< redirection is not available on dash (/bin/sh)
on linux systems, which makes our script and instructions
broken.

Fixes #66

### DIFF
--- a/install-cli.sh
+++ b/install-cli.sh
@@ -28,7 +28,7 @@ validate_checksum() {
 
   checksum=$($checksumbin -a256 "${filename}")
 
-  if grep -Fxq "${checksum}" <<< "${checksumlist}"; then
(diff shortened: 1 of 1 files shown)


cevian (approved):

VineethReddy02:
@blagojts I tested this out and it works.

PR #91, opened 2020-12-14
Add flag to restart pods on pvc expansion

VineethReddy02:
Signed-off-by: Vineeth Pothulapati <vineethpothulapati@outlook.com>

### DIFF
--- a/cli/cmd/volumeExpand.go
+++ b/cli/cmd/volumeExpand.go
@@ -70,11 +70,10 @@ func volumeExpand(cmd *cobra.Command, args []string) error {
 		expandSuccessPrint(pvcPrefix, results)
 
 		if restartsPods {
-			err = k8s.DeletePods(namespace, timescaleDBPodLabels)
+			err = restartPods(timescaleDBPodLabels)
 			if err != nil {
-				return fmt.Errorf("failed to restart pods after PVC expansion: %w", err)
+				return err
 			}
-			fmt.Println("Triggered to restart the pods bound by the PVC's.")
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


cli/cmd/volumeExpand.go
```diff
@@ -68,8 +102,15 @@ func volumeExpand(cmd *cobra.Command, args []string) error {
...
 		}
-
 		expandSuccessPrint(pvcPrefix, map[string]string{pvcPrefix: promStorage})
+
+		if restartsPods {
+			err = k8s.DeletePods(namespace, prometheusPodLabels)
+			if err != nil {
+				return fmt.Errorf("failed to restart pods after PVC expansion: %w", err)
+			}
+			fmt.Println("Triggered to restart the pod bound by the PVC.")
```
blagojts, line 112: duplicate from above, can be extracted to func

cli/pkg/k8s/k8s.go
```diff
@@ -489,6 +489,22 @@ func ExpandPVC(namespace, pvcName, value string) error {
...
+	}
+
+	client, _ := kubeInit()
+	for _, pod := range pods {
+		err = client.CoreV1().Pods(namespace).Delete(context.Background(), pod.Name, metav1.DeleteOptions{})
+		if err != nil {
+			fmt.Printf("failed to delete the pod: %s %v\n", pod.Name, err)
+		}
+	}
+	return nil
```
blagojts, line 505: return an error here if at least one of the pods weren't deleted so the command doesn't terminate with a 0 exit code

blagojts (approved):
Looks good, small comments left, can merge after changes, no need for re-review

PR #93, opened 2020-12-18
README: releases link might give 404

heyleke:
maybe depending on how you reach the README.md, but the releases link is giving a 404 when using the link on "https://github.com/timescale/tobs" page, so replace it with full link

Second part of the diff is caused by doing a 'github' edit I guess, I don't see the diff on that line...


VineethReddy02:
Thanks for your contribution! @heyleke :)

Issue #84, opened 2020-12-02
Make it easy to install with forge using the tobs cli


cevian:
Make it easy to use the hosted platform Forge for running TimescaleDB

PR #100, opened 2021-01-13
tobs install option to connect to an existing db

VineethReddy02:
Support tobs install to connect to an existing database using dbURI. This also supports tobs in connecting with forge. 

Exmaple:

```
tobs install --external-db-uri postgres://<DB_USER>:<DB_PASSWORD>@<DB_HOST>:<DB_PORT>/<DB_NAME>?sslmode=<DB_SSLMODE>
```

Note:

These changes supports `tobs timescaledb` operations such as getPassword, changePassword, etc.. & `tobs metrics` operations such as chunk-interval & retention.  Except for the timescaleDB port-forwarding feature as the DB is outside the k8s cluster. 

This also works in the native helm chart. By modifying few fields in the `values.yaml` i.e.

```
helm install <release-name> timescale/tobs \
--set timescaledb-single.enabled=false,timescaledbExternal.enabled=true,timescaledbExternal.db_uri=<timescaledb-uri>, \
promscale.connection.uri.secretTemplate=<release-name>-timescaledb-uri
```

Resolves: #84


docs/cli-usage.md
```diff
@@ -1,4 +1,4 @@
-## Usage 
+# Usage
```
VineethReddy02, line 1: Formatted this file for better readability.

chart/README.md
```diff
@@ -101,6 +101,8 @@ You can set up the credentials, nodeSelector, volume sizes (default volumes crea
 | `promscale.image`                        | Docker image to use for the Connector                 | `timescale/promscale:0.1.0-alpha.2` |
 | `promscale.connection.dbName`            | Database to store the metrics in                      | `postgres`  |
 | `promscale.connection.user`              | User used for connection to db | `postgres` |
+| `promscale.connection.dbURI.name`         | DB uri name is used as a key for secret which will hold the db URI value | `db-uri` |
+| `promscale.connection.dbURI.secretTemplate` | The template for generating the name of a secret object which will hold the db URI | `` |
```
cevian, line 105: I don't see where these two variables are actually used.

chart/values.yaml
```diff
@@ -153,3 +153,8 @@ promlens:
   defaultPrometheusUrl:  "http://localhost:9201"
   loadBalancer:
     enabled: false
+
+#Enable Promscale to connect to an existing database by providing the db-uri
+external_database:
```
cevian, line 158: i would rename this to `timescaledb-external` and put above timescaledb-single

cli/README.md
```diff
@@ -45,7 +45,7 @@ The following are the commands possible with the CLI.
 
 | Command             | Description                                                      | Flags                                                |
 |---------------------|------------------------------------------------------------------|------------------------------------------------------|
-| `tobs install`      | Alias for `tobs helm install`.                                   | `--filename`, `-f` : file to load configuration from |
+| `tobs install`      | Alias for `tobs helm install`.                                   | `--filename`, `-f` : file to load configuration from <br> `--external-db-uri`, `-e`: external database URI, TimescaleDB installation will be skipped & Promscale connects to the provided database  |
```
cevian, line 48: may be better as --external-timescaledb-uri

cli/pkg/utils/utils.go
```diff
@@ -225,4 +225,67 @@ func ConvertMapI2MapS(v interface{}) interface{} {
...
+type dbDetails struct {
+	DbHost     string
+	DbUser     string
+	DbPort     string
+	DbPassword string
+	SslMode    string
+	DbName     string
+}
+
+func SplitDBURI(dbURI string) (dbDetails, error) {
```
cevian, line 239: Instead of writing this from scratch can we simply use https://github.com/jackc/pgconn/blob/master/config.go#L191

cevian (changes requested):
Some comments

chart/README.md
VineethReddy02: As tobs is a CLI we are assigning these values in `helmInstall.go` I mentioned the fields here because if someone wants to use them while using helm chart. This will be helpful.

chart/values.yaml
VineethReddy02: Having the name with `-` making the chart complex while accessing the fields. [Here](https://github.com/helm/helm/issues/2192) is a discussion on the same in helm repo. So renaming it to camelcase i.e. `timescaledbExternal`

cli/pkg/utils/utils.go
```diff
@@ -225,4 +228,24 @@ func ConvertMapI2MapS(v interface{}) interface{} {
 	}
 
 	return v
-}
\ No newline at end of file
+}
+
+func GetTimescaleDBURI(namespace, name string) (string, error) {
+	dbUriNotFoundError := "secrets \"" + name + "-timescaledb-uri\" not found"
+	dbUriError := errors.New(dbUriNotFoundError)
```
cevian, line 235: is there really no better way to catch this? Can we use a list interface or something?

cevian (changes requested):

chart/templates/external-db-password-secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{ if .Values.timescaledbExternal.enabled -}}
```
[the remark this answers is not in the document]
VineethReddy02: As Grafana Postgres data source plugin needs password & doesn't accept DBURI currently. We need a secret object which holds just a password parsed from db-uri.  So we are maintaining a secret that holds `db_user:db_password` key/value pair.

cli/cmd/helmInstall.go
```diff
@@ -68,7 +76,49 @@ func installStack(file, ref string) error {
...
+				",grafana.timescale.datasource.host="+db.ConnConfig.Host+
+				",grafana.timescale.datasource.port="+port+
+				",grafana.timescale.datasource.dbName="+db.ConnConfig.Database+
+				",grafana.timescale.datasource.sslmode="+sslmode+
+
+				",grafana.timescale.adminUser="+db.ConnConfig.User+
+				",grafana.timescale.adminPassSecret="+name+"-timescaledb-passwords,"+
+
+				"timescaledbExternal.db_password="+db.ConnConfig.Password+
+				",timescaledbExternal.db_user="+db.ConnConfig.User)
```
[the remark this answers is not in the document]
VineethReddy02: I will add them, I was thinking that these fields will only be used by tobs CLI code not by the users directly. But to have tobs chart with more details I will add these fields to the values.yml file.

chart/README.md
```diff
@@ -101,6 +101,7 @@ You can set up the credentials, nodeSelector, volume sizes (default volumes crea
 | `promscale.image`                        | Docker image to use for the Connector                 | `timescale/promscale:0.1.0-alpha.2` |
 | `promscale.connection.dbName`            | Database to store the metrics in                      | `postgres`  |
 | `promscale.connection.user`              | User used for connection to db | `postgres` |
+| `promscale.connection.dbURI.secretTemplate` | The template for generating the name of a secret object which will hold the db URI | `` |
```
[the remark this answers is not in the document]
VineethReddy02: This will be used by Promscale on startup to connect with db-uri.  Here we can provide promscale the db specific details such as host, port, user, password, or the db-uri. It's a matter of preference. But to make db-uri work as intended & considering recent work over support for db-uri in promscale & it's helm chart. I am using db-uri with promscale to make sure it's working as expected.

cli/cmd/timescaledbConnect.go
```diff
@@ -55,6 +60,23 @@ func timescaledbConnect(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("could not get TimescaleDB password: %w", errors.New("user not found"))
 	}
 
+	uri, err := utils.GetTimescaleDBURI(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	if uri != "" {
```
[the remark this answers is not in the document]
VineethReddy02: Parsing the URI made a few changes to the existing flow. So, I preferred to do that by just passing the values to the existing logic. I will update the PR to fully use the URI.

cli/cmd/upgrade.go
```diff
@@ -43,6 +43,11 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("couldn't get the chart-reference flag value: %w", err)
 	}
 
+	dbURI, err := cmd.Flags().GetString("external-timescaledb-uri")
```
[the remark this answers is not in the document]
VineethReddy02: I was initially working on a flow to upgrade tobs installation from existing k8s TimescaleDB to the remote database as forge. But as I went forward it was getting complex in updating too many resources with DB credentials, etc... So I dropped the thought of upgrade support. But somehow missed removing the flag support in upgrade cmd.

chart/templates/external-db-password-secret.yaml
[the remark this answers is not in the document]
VineethReddy02: We update both db-uri password & the normal password on the password change.

cevian (changes requested):
Mostly looks good. two things:
1) there are a lot of unnecessary files in deploy-timescaledb-chart. I actually wonder if it's easier to simply deploy the public timescaledb-single chart than copying over our chart just for timescaledb -- or can't you use our chart but just with a custom values.yml file.

2) Do we need any changes to NOTES.txt for the messaged displayed when using an external URI?

cli/tests/external-db-tests/timescale_test.go
```diff
@@ -0,0 +1,33 @@
...
+		t.Skip("Skipping TimescaleDB tests")
+	}
+
+	fmt.Println("Performing TimescaleDB tests for external db setup...")
+	releaseInfo := test_utils.ReleaseInfo{
+		Release:   RELEASE_NAME,
+		Namespace: NAMESPACE,
+	}
+
+	//releaseInfo.TestTimescaleGetPassword(t, "")
```
cevian, line 20: are these important?

cevian (approved):

cli/tests/external-db-tests/timescale_test.go
VineethReddy02: These are the tests to check change password for TimescaleDB, As mentioned earlier in the test suite as TimescaleDB is in the same cluster. The below call is failing

```
pool, err = pgxpool.Connect(context.Background(), dbURI)
if err != nil {
  return nil, err
}
```

Error:

```
Error: could not change TimescaleDB password: failed to connect to `host=external-db-tests.default.svc.cluster.local user=postgres database=postgres`: hostname resolving error (lookup external-db-tests.default.svc.cluster.local: no such host)
```

This works with forge. but some issue with the local TImescale DNS resolution.

PR #103, opened 2021-01-27
Fixed extraScrapeConfigs: parameter example in values.yaml

sokoow:
extraScrapeConfigs field in values.yaml actually expects a string - fixed that and also added a small example of using it


VineethReddy02:
Thanks, @sokoow for your contribution. This LGTM!

PR #109, opened 2021-02-19
refactor helm, add external db tests & fix few issues

VineethReddy02:
1. Refactor helm around external db uri logic by adding template functions.
2. Add tests to verify change password when tobs is connected with external db by creating NodePort service to timescaleDB. 
3. Cleaned up e2e tests. 
4. Fixed some silent test failures.
5. Add additional tests to verify that all pods are in the running state that are deployed.


cli/cmd/grafanaChangePassword.go
```diff
@@ -45,10 +45,27 @@ func grafanaChangePassword(cmd *cobra.Command, args []string) error {
 
 	err = k8s.KubeExecCmd(namespace, grafanaPod, "grafana", "grafana-cli admin reset-admin-password "+password, nil, false)
 	if err != nil {
-		secret.Data["admin-password"] = oldpassword
-		_ = k8s.KubeUpdateSecret(namespace, secret)
-		return fmt.Errorf("could not change Grafana password: %w", err)
+		err1 := updateToOldPassword(oldpassword)
```
VineethReddy02, line 48: Noticed an issue in this use-case, If the new password is short exmaple `hii` pwd updation fails from grafana stating pwd is too short. But here our fallback mechanism is trying to update back to the old pwd by updating the secret object we hold from the previous get operation, This is retrieved before updating the new password. 

This updation fails as the secret object we are trying to assign is too old and metadata of the secret doesn't let us use the object to update pwd. So we need to retrieve again the pwd for updataion.

cevian (approved):

PR #112, opened 2021-03-04
add experimental support for kube-prometheus stack

VineethReddy02:
Below cmd will install Kube-Prometheus in tobs and will disable default Prometheus & Grafana deployments as Kube-Prometheus will deploy those components. 

```
tobs install -c ./chart/ --enable-kube-prometheus
```

**Note**: This PR is dependent on #114 as there is some common functionality & common refactoring. This PR is rebased on top of that PR. 

Todos:

1. Documents the Kube-Prometheus add-on. 
2. Add new tobs CLI commands to manage & configure Kube-Prometheus.
3. Add an option to deploy Kube-Prometheus with Prometheus HA.


chart/values.yaml
```diff
@@ -159,6 +159,22 @@ promlens:
...
 
+kube-prometheus-stack:
+  enabled: false
+  prometheus:
+    prometheusSpec:
+      ## The remote_read spec configuration for Prometheus.
+      ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#remotereadspec
+      remoteRead:
+      # - {protocol}://{host}:{port}/{endpoint}
+        - url: "http://tobs-promscale-connector.default.svc.cluster.local:9201/read"
```
blagojts, line 170: can we use the same template that the prom part of the chart uses?

cli/cmd/helmInstall.go
```diff
@@ -50,14 +51,19 @@ func helmInstall(cmd *cobra.Command, args []string) error {
...
+
+	err = installStack(file, ref, dbURI, kp)
 	if err != nil {
 		return fmt.Errorf("could not install The Observability Stack: %w", err)
 	}
 	return nil
 }
 
-func installStack(file, ref, dbURI string) error {
+func installStack(file, ref, dbURI string, kp bool) error {
```
blagojts, line 66:
```suggestion
func installStack(file, ref, dbURI string, enableKubePrometheus bool) error {
```

cli/cmd/helmInstall.go
```diff
@@ -119,11 +132,13 @@ func installStack(file, ref, dbURI string) error {
...
 
-func appendDBURIValues(dbURI, name string, cmds []string) ([]string, error) {
-	cmds = append(cmds, "--set",
-		"timescaledb-single.enabled=false,"+
-			"timescaledbExternal.enabled=true,"+
-			"timescaledbExternal.db_uri="+dbURI+
-			",promscale.connection.uri.secretTemplate="+name+"-timescaledb-uri,")
-	return cmds, nil
+func appendDBURIValues(dbURI, name string, helmValues string) string {
+	helmValues = helmValues+",timescaledb-single.enabled=false,"+"timescaledbExternal.enabled=true,"+"timescaledbExternal.db_uri="+dbURI+
```
blagojts, line 136: tl;dr

blagojts (changes requested):

chart/values.yaml
```diff
@@ -159,6 +159,22 @@ promlens:
   loadBalancer:
     enabled: false
```
cevian, line 161: We should put some comments about this being experimental and also highlight which components should be disabled if this is enabled. In fact, I might have a kube_prometheus_values.yml file in this dir that would serve as an example of what that would look like

cevian (changes requested):
Overall looks good, but I think we need much better documentation

cevian (commented):
should we have tobs uninstall kube-prometheus crds automatically https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#uninstall-chart

cli/cmd/helmInstall.go
VineethReddy02: This refactoring of existing logic. When connecting to external DB. Tobs needs to disabled TimescaleDB installation and map values to timescaleDBexternal fields accordingly.

PR #114, opened 2021-03-12
Update dependent helm chart + code refactoring

VineethReddy02:
1. The default user for timecaledb is `postgres` as the latest timescaledb helm chart doesn't hold db secret user as a key for the db password. We should parse user from `postgres` to `PATRONI_SUPERUSER_PASSWORD` which is hardcoded in timescaleDB secret.
2. The new timescaledb helm chart expects secrets to be created before installing the timescaleDB. So tobs will create `<RELEASE_NAME-credentials>`, `RELEASE_NAME-certificate` secrets prior to install step.
3. Added a new flag `--enable-timescaledb-backup` or `b` for `tobs install` & `tobs upgrade` this step will request the user to provide the s3 details. If expected details are configured to env variables we will skip the user prompts for the details. This also creates `RELEASE_NAME-pgbackrest` secret with provided details
4. The default timescaleDB secret name has been changed from `<RELEASE_NAME-timescaledb-passwords>` to `<RELEASE_NAME-credentials>`. 
5. Refactored helm chart templating functions to drop redundant logic from `grafana-db-user-jonbyaml`
6. The default timescaleDB user will be fetched from helm reelase i.e. `timescaledb-single.patroni.postgresql.authentication.superuser.username` the default secret password will be fetched using `PATRONI_SUPERUSER_PASSWORD` key.

Few refactorings:
1. Changed the `name` global variable to `helmReleaseName` this widely used across go files. 
2. `tobs timescaledb connect` spins up a `psql` pod to create a SPDY session. When we exit from psql the pod isn't being deleted previously. Refactored that part of the logic to terminate immediately by setting `gradePeriodSeconds as 0`.


VineethReddy02:
Note: CI fails as the latest changes expect Promscale support them. As the recent changes made in Promscale helm chart aren't released yet. So the CI fails.

cli/pkg/pgconn/pgconn.go
```diff
@@ -63,6 +63,10 @@ func OpenConnectionToDB(namespace, name, user, dbname string, remote int) (*pgxp
 		return nil, err
 	}
 
+	if user == "PATRONI_SUPERUSER_PASSWORD" {
+		user = "postgres"
+	}
+
```
atanasovskib, line 69: I mentioned this above, I really don't like this

cli/pkg/timescaledb_secrets/create_secrets.go
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		fmt.Println("skipping to create TimescaleDB s3 backup secret as backup option is disabled.")
+	}
+
+	return nil
+}
+
+func createS3BackupForTimescaleDB(releaseName, namespace string) error {
+	s3 := k8s.S3Details{}
+	fmt.Print("We'll be asking a few questions about S3 buckets, keys, secrets and endpoints.\n\nFor background information, visit these pages:\n\nAmazon Web Services:\n- https://docs.aws.amazon.com/AmazonS3/latest/gsg/CreatingABucket.html\n- https://docs.aws.amazon.com/IAM/latest/UserGuide/introduction.html\n\nDigital Ocean:\n- https://developers.digitalocean.com/documentation/spaces/#aws-s3-compatibility\n\nGoogle Cloud:\n- https://cloud.google.com/storage/docs/migrating#migration-simple\n\n")
```
atanasovskib, line 43: Use raw string literals 
```
`here
there`
```

cli/pkg/timescaledb_secrets/create_secrets.go
```diff
@@ -0,0 +1,84 @@
...
+		fmt.Scanln(&s3.EndpointName)
+	}
+
+	fmt.Println("\nWhat is the region of the S3 endpoint? (leave blank for default)")
+	s3.EndpointRegion = os.Getenv("PGBACKREST_REPO1_S3_REGION")
+	if s3.EndpointRegion == "" {
+		fmt.Scanln(&s3.EndpointRegion)
+	}
+
+	fmt.Println("\nWhat is the S3 Key to use?")
```
atanasovskib, line 65: Modify prompts, ask "What" only if env not set, prompt "S3 key found in env, using"

cli/pkg/utils/utils.go
```diff
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 
+	yaml2 "sigs.k8s.io/yaml"
```
atanasovskib, line 14: why two versions?

cli/pkg/utils/utils.go
```diff
@@ -243,10 +245,53 @@ func GetTimescaleDBURI(namespace, name string) (string, error) {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to parse values.yaml to json bytes %v", err)
+	}
+
+	// JSON object parses into a map with string keys
+	itemsMap := f.(map[string]interface{})
+
+	// Loop through the Items; we're not interested in the key, just the values
+	for k, v := range itemsMap {
+		if k == "timescaledb-single" {
```
atanasovskib, line 280:
```suggestion
		if k != "timescaledb-single" {
		  continue
		}
```

More readable when we reduce nesting

cli/pkg/utils/utils.go
```diff
@@ -243,10 +245,53 @@ func GetTimescaleDBURI(namespace, name string) (string, error) {
...
+			// Access the values in the JSON object and place them in an Item
+			for itemKey, itemValue := range v.(map[string]interface{}) {
+				if itemKey == "backup" {
```
atanasovskib, line 283: samesies as above

cli/tests/testdata/chart1/values.yaml
```diff
@@ -3,17 +3,21 @@
 #    https://github.com/timescale/timescaledb-kubernetes/tree/master/charts/timescaledb-single
```
atanasovskib, line 3: Not for this PR, but these need to be refactored really really bad

atanasovskib (changes requested):

cli/tests/testdata/chart1/values.yaml
VineethReddy02: Can you elaborate on this? What needs to be changed here.

chart/values.yaml
```diff
@@ -8,12 +8,16 @@ timescaledb-single:
   # disable the chart if an existing TimescaleDB instance is used
   enabled: true
   image:
-    tag: pg12-ts1.7-latest
+    tag: pg12-ts2.0-latest
```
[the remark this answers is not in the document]
VineethReddy02: We added support in Promscale to support all TimescaleDB 2.x versions i.e. `>=1.7.3 <2.99.0` so the image should only hold all 2.x latest versions as the TimescaleDB progresses. But I am okay with your suggestion. 

Does this mean on every tobs release we should update the timescaleDB image manually based on Compatability?

cli/cmd/timescaledbChangePassword.go
```diff
@@ -22,7 +22,7 @@ var timescaledbChangePasswordCmd = &cobra.Command{
 
 func init() {
 	timescaledbCmd.AddCommand(timescaledbChangePasswordCmd)
-	timescaledbChangePasswordCmd.Flags().StringP("user", "U", "postgres", "user whose password to change")
+	timescaledbChangePasswordCmd.Flags().StringP("user", "U", "PATRONI_SUPERUSER_PASSWORD", "user whose password to change")
```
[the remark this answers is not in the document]
VineethReddy02: We expect them to specify the `user` & if there is a new user in DB we expect the password to the user to be part of the default TimescaleDB credential secret object. But for default postgres user the key in secret is `PATRONI_SUPERUSER_PASSWORD` we will use it as a secret key reference while dealing with postgres user operations.

cli/tests/testdata/chart1/values.yaml
atanasovskib: Replied in DM :) 
This way is very tedious and error prone, we should just reference the helm chart directly and have only the changed value specified in the tests. Not to be done in this PR

cli/cmd/timescaledbConnect.go
```diff
@@ -26,7 +26,7 @@ var timescaledbConnectCmd = &cobra.Command{
 
 func init() {
 	timescaledbCmd.AddCommand(timescaledbConnectCmd)
-	timescaledbConnectCmd.Flags().StringP("user", "U", "PATRONI_SUPERUSER_PASSWORD", "database user name")
+	timescaledbConnectCmd.Flags().StringP("user", "U", "", "database user name, if not provided will be used super-user used for deployment")
```
atanasovskib, line 29:
```suggestion
	timescaledbConnectCmd.Flags().StringP("user", "U", "", "database user name, if not provided super-user will be used for deployment")
```

cli/cmd/chunkIntervalGet.go
```diff
@@ -27,7 +29,20 @@ func chunkIntervalGet(cmd *cobra.Command, args []string) error {
 
 	metric := args[0]
 
-	pool, err := pgconn.OpenConnectionToDB(namespace, name, user, dbname, FORWARD_PORT_TSDB)
+	secretKey, user, err := utils.GetDBSecretKeyAndDBUser(helmReleaseName, user)
+	if err != nil {
+		return fmt.Errorf("could not get DB secret key from gelm release: %w", err)
```
atanasovskib, line 34:
```suggestion
		return fmt.Errorf("could not get DB secret key from helm release: %w", err)
```

saw gelm a few times, I will only leave one comment

cli/cmd/chunkIntervalReset.go
```diff
@@ -26,7 +28,20 @@ func chunkIntervalReset(cmd *cobra.Command, args []string) error {
...
+	}
+
+	d := pgconn.DBDetails{
+		Namespace: namespace,
+		Name:      helmReleaseName,
+		DBName:    dbname,
+		User:      user,
+		SecretKey: secretKey,
+		Remote:    FORWARD_PORT_TSDB,
+	}
```
atanasovskib, line 43: I think there's a possibility to abstract these lines in a function, I think they are the same throughout the code (e.g. here and chunkIntervalGet.go lines 32-44)

cli/cmd/chunkIntervalSet.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/timescale/tobs/cli/pkg/utils"
+
```
atanasovskib, line 10: extra line not needed

cli/cmd/timescaledbChangePassword.go
```diff
@@ -22,7 +22,7 @@ var timescaledbChangePasswordCmd = &cobra.Command{
 
 func init() {
 	timescaledbCmd.AddCommand(timescaledbChangePasswordCmd)
-	timescaledbChangePasswordCmd.Flags().StringP("user", "U", "PATRONI_SUPERUSER_PASSWORD", "user whose password to change")
+	timescaledbChangePasswordCmd.Flags().StringP("user", "U", "", "user whose password to change, if not provided will be used super-user used for deployment")
```
atanasovskib, line 25:
```suggestion
	timescaledbChangePasswordCmd.Flags().StringP("user", "U", "", "user whose password will be changed, if not provided super-user's password will be changed")
```

cli/cmd/timescaledbGetPassword.go
```diff
@@ -19,7 +19,7 @@ var timescaledbGetPasswordCmd = &cobra.Command{
 
 func init() {
 	timescaledbCmd.AddCommand(timescaledbGetPasswordCmd)
-	timescaledbGetPasswordCmd.Flags().StringP("user", "U", "PATRONI_SUPERUSER_PASSWORD", "user whose password to get")
+	timescaledbGetPasswordCmd.Flags().StringP("user", "U", "", "user whose password to get, if not provided will be used super-user used for deployment.")
```
atanasovskib, line 22: `, if not provided will be used super-user used for deployment.` is used several times where a `user` flag is defined for a command, but:
1. can be reworded as `if not provided super-user will be used for deployment`
2. we're not always doing a deployment in some of these commands, we need a diff msg there (i.e. here)

cli/pkg/pgconn/pgconn.go
```diff
@@ -79,8 +78,10 @@ func OpenConnectionToDB(namespace, name, user, dbname string, remote int) (*pgxp
 		}
 		local := int(ports[0].Local)
 
-		pool, err = pgxpool.Connect(context.Background(), "postgres://"+user+":"+pass+"@localhost:"+strconv.Itoa(local)+"/"+dbname)
+		pool, err = pgxpool.Connect(context.Background(), "postgres://"+d.User+":"+pass+"@localhost:"+strconv.Itoa(local)+"/"+d.DBName)
```
atanasovskib, line 81: fmt.Sprintf

cli/pkg/pgconn/pgconn.go
```diff
@@ -79,8 +78,10 @@ func OpenConnectionToDB(namespace, name, user, dbname string, remote int) (*pgxp
...
 		if err != nil {
+			fmt.Println(pass)
+			fmt.Println("HIII")
```
atanasovskib, line 84: hi back

atanasovskib (approved):
Minor nits over formatting, overall changes look good

cli/pkg/timescaledb_secrets/create_secrets.go
```diff
@@ -0,0 +1,109 @@
...
+Digital Ocean:
+- https://developers.digitalocean.com/documentation/spaces/#aws-s3-compatibility
+
+Google Cloud:
+- https://cloud.google.com/storage/docs/migrating#migration-simple
+
+`)
+	// if values are available in env variables pick up them else
+	// read from the user input.
+	s3.BucketName = os.Getenv("PGBACKREST_REPO1_S3_BUCKET")
```
cevian, line 62: Are these environmental variable names documented anywhere by us? What do you think of documenting them either in the prompt text or in the values.yml (next to the backup enabled section)?

cli/pkg/k8s/k8s.go
```diff
@@ -518,4 +526,239 @@ func DeletePods(namespace string, labels map[string]string) error {
...
+func DeleteTimescaleDBSecrets(releaseName, namespace string, backUpEnabled bool) {
+	client, _ := kubeInit()
+	fmt.Println("Deleting TimescaleDB secrets...")
+	credentialsSecret := []string{releaseName + "-credentials", releaseName + "-certificate"}
+	if backUpEnabled {
+		credentialsSecret = append(credentialsSecret, releaseName+"-pgbackrest")
+	}
+	for _, s := range credentialsSecret {
+		err := client.CoreV1().Secrets(namespace).Delete(context.Background(), s, metav1.DeleteOptions{})
+		if err != nil {
```
cevian, line 760: let's check https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound and ignore error if not found

cli/pkg/utils/utils.go
```diff
@@ -243,10 +243,114 @@ func GetTimescaleDBURI(namespace, name string) (string, error) {
...
+
+	r, err := findKeysFromYaml(res,keys)
+	if err != nil {
+		return nil, err
+	}
+
+	return r, nil
+}
+
+func ExportValuesFieldFromReleae(releaseName string, keys []string) (interface{}, error) {
```
cevian, line 269:
```suggestion
func ExportValuesFieldFromRelease(releaseName string, keys []string) (interface{}, error) {
```

cli/cmd/helmUninstall.go
```diff
@@ -68,19 +70,26 @@ func helmUninstall(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		fmt.Println(err, ", skipping")
 	}
 
-	err = k8s.KubeDeleteEndpoint(namespace, name)
+	e, err := utils.ExportValuesFieldFromReleae(helmReleaseName, TimescaleDBBackUpKeyForValuesYaml)
+	enableBackUp := e.(bool)
+	if err != nil {
+		return err
+	}
```
cevian, line 82:
```suggestion
	e, err := utils.ExportValuesFieldFromReleae(helmReleaseName, TimescaleDBBackUpKeyForValuesYaml)
        if err != nil {
		return err
	}
	enableBackUp, ok := e.(bool)
	if !ok {
	  return fmt.Errorf("Enable Backup was not a bool")
	}

```

cevian: casting without an ok is dangerous as it will panic

cli/cmd/helmInstall.go
```diff
@@ -92,6 +103,26 @@ func installStack(file, ref, dbURI string) error {
 		cmds = append(cmds, "--devel")
 	}
 
+	// If enable backup is disabled by flag check the backup option
+	// from values.yaml as a second option
+	if !enableBackUp {
+		e, err := utils.ExportValuesFieldFromChart(ref, TimescaleDBBackUpKeyForValuesYaml)
+		enableBackUp = e.(bool)
```
cevian, line 110: check if the cast to bool is ok, do so only after checking the err

cevian (changes requested):
Some changes requested. The most important ones being around security and the generation of passwords and certificates.

cli/pkg/utils/utils.go
```diff
@@ -243,10 +243,114 @@ func GetTimescaleDBURI(namespace, name string) (string, error) {
...
+		dbUser = fmt.Sprintf("%v", userName)
+		// As the user isn't provided
+		// use default super user from helm release
+		// the default super-user password is mapped to "PATRONI_SUPERUSER_PASSWORD" secret key
+		secretKey = "PATRONI_SUPERUSER_PASSWORD"
+	} else {
+		secretKey = dbUser
+	}
+
+	return secretKey, dbUser, nil
```
antekresic, line 355:
```suggestion
	if dbUser != "" {
		return dbUser, dbUser, nil
	}

	userName, err := ExportValuesFieldFromReleae(releaseName, []string{"timescaledb-single", "patroni", "postgresql", "authentication", "superuser", "username"})
	if err != nil {
		return "", "", err
	}
	
	// As the user isn't provided
	// use default super user from helm release
	// the default super-user password is mapped to "PATRONI_SUPERUSER_PASSWORD" secret key
	return "PATRONI_SUPERUSER_PASSWORD", userName, nil
```

antekresic: Might want to wrap that error too.

cli/pkg/utils/utils_test.go
```diff
@@ -0,0 +1,68 @@
...
+		wantErr bool
+	}{
+		{
+			name: "Success test-case to fetch nothing as the provided keys are empty",
+			args: args{
+				chart: "../../../chart/",
+				keys:  nil,
+			},
+			want: nil,
+			wantErr: false,
```
antekresic, line 26: I would omit `wantErr` from here since by default its false, easier to scan for tests which return errors.

I would personally do the same for `want` since default is nil.

cli/pkg/pgconn/pgconn.go
```diff
@@ -23,11 +31,18 @@ func OpenConnectionToDB(namespace, name, user, dbname string, remote int) (*pgxp
...
-	tspromPods, err := k8s.KubeGetPods(namespace, map[string]string{"app": name + "-promscale"})
+	tspromPods, err := k8s.KubeGetPods(d.Namespace, map[string]string{"app": d.Name + "-promscale"})
 	if err != nil {
 		return nil, err
 	}
 
+	passBytes, err := utils.GetDBPassword(d.SecretKey, d.Name, d.Namespace)
+	if err != nil {
+		fmt.Println(err)
+		return nil, err
```
antekresic, line 42: This seems like an anti-pattern. Printing an error (I should say logging it to stdout) and then returning it, where it will most likely be logged later on in the stack.

antekresic (changes requested):
Plenty of updates necessary. 

Might be out of scope for this PR but a logging system is needed to replace the abundance of `fmt.Print` calls everywhere.

cli/pkg/k8s/k8s.go
```diff
@@ -518,4 +525,234 @@ func DeletePods(namespace string, labels map[string]string) error {
...
+func checkSecretExists(secretName, namespace string) (bool, error) {
+	client, _ := kubeInit()
+	secExists, err := client.CoreV1().Secrets(namespace).Get(context.Background(), secretName, metav1.GetOptions{})
+	e1 := errors.New("secrets \"" + secretName + "\" not found")
+	if err != nil && err.Error() != e1.Error() {
+		return false, fmt.Errorf("failed to get %s secret present or not %v", secretName, err)
+	}
+
+	if secExists.Name != "" {
+		fmt.Printf("using existing %s secret\n", secretName)
```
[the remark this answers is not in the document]
VineethReddy02: I will move this out to the calling fuc.

cli/pkg/k8s/k8s.go
```diff
@@ -518,4 +525,234 @@ func DeletePods(namespace string, labels map[string]string) error {
...
+	exists, err := checkSecretExists(secretName, namespace)
+	if err != nil {
+		return err
+	}
+
+	if exists {
+		return nil
+	}
+
+	fmt.Printf("Creating TimescaleDB %s secret\n", secretName)
```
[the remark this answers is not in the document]
VineethReddy02: I will move all the action specific messages to the caller from k8s.

cli/pkg/timescaledb_secrets/create_secrets.go
VineethReddy02: Added the docs in `values.yaml` just above the `backup` option.

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,316 @@
...
+// Compatible with passwords generated by the postgres operator
+func randomPassword(n int) ([]byte, error) {
+	var passwordChars = []byte("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
+	b := make([]byte, n)
+	_, err := crand.Read(b)
+	if err != nil {
+		return b, fmt.Errorf("failed to generate random password %v", err)
+	}
+	for i := 0; i < n; i++ {
+		b[i] = passwordChars[b[i]/5]
```
cevian, line 74: this should read `b[i] = passwordChars[uint8(b[i]) % len(passwordChars)]`

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate serial number: %v", err)
+	}
+
+	template := x509.Certificate{
+		SerialNumber:          serialNumber,
```
cevian, line 133: let's add 	following the kubernetes ingress controller https://i.stack.imgur.com/iyjjx.png	
Subject: pkix.Name{
			CommonName:   "Tobs TimescaleDB Fake Certificate",
			Organization: []string{"Acme Co"},
},

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,316 @@
...
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
+
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create certificate: %v", err)
+	}
+
+	certOut := new(bytes.Buffer)
```
cevian, line 146: can we copy over `PEMEncodeCert(certDerBytes []byte) ([]byte, error)` and use that

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,316 @@
...
+	if err := pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes}); err != nil {
+		return nil, nil, fmt.Errorf("failed to write data to cert.pem: %v", err)
+	}
+
+	privBytes, err := x509.MarshalECPrivateKey(priv)
```
cevian, line 151: can we copy over  `PEMEncodePrivateKey(privateKey *ecdsa.PrivateKey) ([]byte, error)` and use that?

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,316 @@
...
+	BucketName     string
+	EndpointName   string
+	EndpointRegion string
+	Key            string
+	Secret         string
+}
+
+func (t *TSDBSecretsInfo) createTimescaleDBCertificates() error {
+	fmt.Print(`
+WARNING: The generated certificate in this directory is self-signed and is only
```
cevian, line 174: Lets replace with:

```
WARNING: Using a generated self-signed certificate for TLS access to TimescaleDB.
                    This should only be used for development and demonstration purposes.
                    To use a signed certificate, use the `--tls-timescaledb-cert` and `--tls-timescaledb-key`
                    flags when issuing the tobs install command.
```

cevian (changes requested):

cli/cmd/helm/helmInstall.go
```diff
@@ -38,42 +35,74 @@ func addHelmInstallFlags(cmd *cobra.Command) {
...
 		return fmt.Errorf("could not install The Observability Stack: %w", err)
 	}
 
-	err = installStack(file, ref, dbURI, enableBackUp)
+	keyFile, err := cmd.Flags().GetString("timescaledb-tls-key")
+	if err != nil {
+		return fmt.Errorf("could not install The Observability Stack: %w", err)
+	}
+
+	if certFile != "" && keyFile != "" {
```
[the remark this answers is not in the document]
VineethReddy02: We need both to be provided right? If not none. We cannot take one value from the user & the other from our generation. So if  both are not empty then we proceed to read them. If not we generate them ourseleves.

cli/cmd/metrics/retentionSet.go
```diff
@@ -29,18 +30,18 @@ func retentionSet(cmd *cobra.Command, args []string) error {
...
 	d := pgconn.DBDetails{
-		Namespace: namespace,
-		Name:      helmReleaseName,
+		Namespace: root.Namespace,
+		Name:      root.HelmReleaseName,
 		DBName:    dbname,
 		User:      user,
 		SecretKey: secretKey,
-		Remote:    FORWARD_PORT_TSDB,
+		Remote:    5432,
```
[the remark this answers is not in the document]
VineethReddy02: This isn't relevant anymore. Not sure how you are able to see them in the latest files...

cli/pkg/timescaledb_secrets/create.go
VineethReddy02: Updating it to this `b[i] = passwordChars[b[i] % uint8(len(passwordChars))]` as `uint8(b[i])` is redundant conversion.

cli/pkg/timescaledb_secrets/create.go
VineethReddy02: It's the same, I compared it. But simplicity will copy it here.

cli/cmd/helm/helmInstall.go
```diff
@@ -0,0 +1,202 @@
...
+	install := exec.Command("helm", cmds...)
+	fmt.Println("Installing The Observability Stack")
+	out, err := install.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not install The Observability Stack: %w \nOutput: %v", err, string(out))
+	}
+
+	time.Sleep(10 * time.Second)
+
+	fmt.Println("Waiting for pods to initialize...")
```
antekresic, line 180: Should this `Println` be before the `Sleep`?

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,344 @@
+package timescaledb_secrets
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	crand "crypto/rand"
```
antekresic, line 8: Why two imports to the same lib? Am I missing something?

cli/pkg/timescaledb_secrets/create.go
```diff
@@ -0,0 +1,344 @@
...
+	Key            string
+	Secret         string
+}
+
+func (t *TSDBSecretsInfo) createTimescaleDBCertificates() error {
+	fmt.Print(`
+WARNING: Using a generated self-signed certificate for TLS access to TimescaleDB.
+         This should only be used for development and demonstration purposes.
+         To use a signed certificate, use the "--tls-timescaledb-cert" and "--tls-timescaledb-key"
+         flags when issuing the tobs install command.
```
antekresic, line 177: I would personally add a check after this message, something like `Do you wish to proceed? [y/n]`. 

I haven't seen this behavior anywhere in TOBS though, I would be fine to add it in a later PR if we decide to do it at all.

antekresic (approved):
Couple of questions/suggestions but nothing major.

The code is looking OK, there are a lot of candidates for basic refactoring but lets get back to that in a future PR.

cli/pkg/timescaledb_secrets/create.go
VineethReddy02: So basically we avoided that because we want one cmd to install tobs without any user inputs. To make it as easy as possible.

cli/cmd/helm/helmInstall.go
VineethReddy02: So the install includes two steps 
1. tobs install internally triggers helm install process. Here we are waiting `10 secs` after helm install to proceed with the step 2. 
2. Check for pods status. After the successful helm install. tobs waits for pods to get to running state by polling/waiting for pods to get into running state before the `tobs install` cmd exits. 

So the wait is to let `helm install` cmd to kick off. So that we can check the pods status that needs to be running after helm install.

cli/pkg/timescaledb_secrets/create.go
cevian: I think this is ok for now. It's a pretty fat warning

cli/cmd/helm/helmInstall.go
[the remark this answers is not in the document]
cevian: Yes, but if a user provieded one and not the other, it's some kind of user error and we should fail with an error message of some sort.

cevian (approved):

PR #118, opened 2021-04-07
upgrade path feature for TimescaleDB 0.5.0 to 0.8.2

VineethReddy02:


cli/cmd/helm/upgrade.go
```diff
@@ -179,3 +188,187 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
...
+	}
+
+	dVersion, err := utils.ParseVersion(deployedChartVersion, 3)
+	if err != nil {
+		return fmt.Errorf("failed to parse deployed helm chart version %w", err)
+	}
+
+	fmt.Println(deployedChartVersion, nVersion, dVersion)
+	switch {
+	case deployedChartVersion <= "0.2.2" && nVersion > dVersion:
```
cevian, line 205: these comparisons aren't right, you can't use string comparison. also I think you want nVersion > cutVersion not nVersion > dVsersion
```suggestion
	cutVersion, err := utils.ParseVersion("0.2.2", 3)
	if err != nil {
		return fmt.Errorf("failed to parse deployed helm chart version %w", err)
	}
	fmt.Println(deployedChartVersion, nVersion, dVersion)
	switch {
	case dVersion <= cutVersion && nVersion > cutVersion:
```

cli/cmd/helm/upgrade.go
```diff
@@ -179,3 +188,187 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse latest helm chart version %w", err)
+	}
+
+	dVersion, err := utils.ParseVersion(deployedChartVersion, 3)
+	if err != nil {
+		return fmt.Errorf("failed to parse deployed helm chart version %w", err)
+	}
+
+	fmt.Println(deployedChartVersion, nVersion, dVersion)
```
cevian, line 203: get rid of this

cli/cmd/helm/upgrade.go
```diff
@@ -179,3 +188,187 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
...
+	fmt.Println("Migrating backup configuration")
+
+	newData := make(map[string][]byte)
+	for key, value := range s.Data {
+		ok := strings.Contains(key, "-s3-")
+		if ok {
+			newKey := strings.Replace(key, "-", "_", -1)
+			newKey = strings.ToUpper(newKey)
+			newKey = "PGBACKREST_" + newKey
+			newData[newKey] = value
```
cevian, line 352: should the value be base64 encoded?

VineethReddy02: encoding is done internally. We just pass the data in the format of bytes.

cli/cmd/helm/upgrade.go
VineethReddy02: So here the condition is
`dVersion <= version0_2_2 && nVersion > dVersion` so I expect the deployed version to be <= 0.2.2 and new Version to be greater than deployed Version.`

The deployed version in users env can be 0`.1.1` too.

cevian (approved):

PR #122, opened 2021-04-14
optimise CI

VineethReddy02:


cli/cmd/helm/helmInstall.go
```diff
@@ -37,13 +37,19 @@ func addHelmInstallFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolP("enable-timescaledb-backup", "b", false, "Enable TimescaleDB S3 backup")
 	cmd.Flags().StringP("timescaledb-tls-cert", "", "", "Option to provide your own tls certificate for TimescaleDB")
 	cmd.Flags().StringP("timescaledb-tls-key", "", "", "Option to provide your own tls key for TimescaleDB")
+	cmd.Flags().StringP("version", "", "", "Option to provide your tobs helm chart version, if not provided will install the latest tobs chart available")
+	cmd.Flags().BoolP("only-secrets", "", false, "Option to create only TimescaleDB secrets")
```
cevian, line 41: These two options should really have been a separate PR or commit. Ok for now, but lets be more careful in future

cli/cmd/helm/helmInstall.go
```diff
@@ -104,6 +122,27 @@ func helmInstall(cmd *cobra.Command, args []string) error {
...
+		helmValues = appendDBURIValues(c.dbURI, cmd.HelmReleaseName, helmValues)
+	} else {
+		// if db-uri is provided we do not need
+		// to create DB level secrets
+		err = c.createSecrets()
+		if err != nil {
+			return fmt.Errorf("failed to create secrets %v", err)
+		}
+		if c.onlySecrets {
+			fmt.Println("Skipping tobs installation.")
```
cevian, line 140:
```suggestion
			fmt.Println("Skipping tobs installation because of only-secrets flag.")
```

cli/cmd/helm/upgrade.go
```diff
@@ -182,7 +182,8 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
 	out := exec.Command("helm", cmds...)
 	k, err := out.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("failed to upgrade %v", string(k))
+		fmt.Println(string(k))
```
cevian, line 185: should this be removed?

cli/cmd/helm/upgrade.go
```diff
@@ -182,7 +182,8 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("failed to upgrade %v", err)
```
cevian, line 186:
```suggestion
		return fmt.Errorf("failed to upgrade %w", err)
```

cli/tests/tobs-cli-tests/installation_test.go
```diff
@@ -171,39 +178,51 @@ func testHelmShowValues(t testing.TB) {
 }
 
 func TestInstallation(t *testing.T) {
-	if testing.Short() {
+	v := os.Getenv("SKIP_INSTALL_TESTS")
+
+	if testing.Short() || v != "" {
```
cevian, line 183: we should be conservative to make sure we don't skip tests
```suggestion
	if testing.Short() || v == "TRUE" {
```

cli/tests/upgrade-tests/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+		log.Fatal(err)
+	}
+
+	os.Exit(code)
+}
+
+func installTobsRecentRelease() {
+	// Note: The below tobs cmd only deploys TimescaleDB as the test values.yaml is configured
+	// to deploy only timescaleDB
+	log.Println("Installing Tobs 0.2.2")
```
cevian, line 53: 0.2.2 should be defiined in a const variable

cli/tests/upgrade-tests/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+	err := test_utils.CheckPodsRunning(NAMESPACE)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	os.Exit(code)
+}
+
+func installTobsRecentRelease() {
+	// Note: The below tobs cmd only deploys TimescaleDB as the test values.yaml is configured
```
cevian, line 51: why are we doing only timescaleDB here?

cli/tests/upgrade-tests/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+	signal.Notify(sigchan, os.Interrupt)
+	go func() {
+		<-sigchan
+		done <- true
+		os.Exit(1)
+	}()
+
+	installTobsRecentRelease()
+
+	upgradeTobsLatest()
```
cevian, line 33: Not sure why these are in TestMain and not in a test run by m.Run?

cli/cmd/helm/helmInstall.go
```diff
@@ -205,3 +229,36 @@ func appendDBURIValues(dbURI, name string, helmValues string) string {
...
+	var i int64
+	var err error
+	if c.version != "" {
+		i, err = utils.ParseVersion(c.version, 3)
+		if err != nil {
+			return fmt.Errorf("failed to parse version %s %v", c.version, err)
+		}
+	}
+
+	// here 3000 represent version
```
cevian, line 243: not for this PR but we should switch a proper semver parsing lib

cli/cmd/helm/helmInstall.go
```diff
@@ -104,6 +122,27 @@ func helmInstall(cmd *cobra.Command, args []string) error {
 
 func (c *installSpec) installStack() error {
 	var err error
+
+	// Note do not change the below order the --set flag is set
```
cevian, line 126: I'm not sure what order this is referring to anymore

cevian (changes requested):
Looks good but I had some questions

cli/cmd/helm/helmInstall.go
VineethReddy02: `only-secrets` I agree. But `version` is added as a utility for upgrade tests. We use `--version` to install `0.2.2` and to upgrade to the latest. So added it as a part of this PR.

cli/cmd/helm/helmInstall.go
VineethReddy02: The comment line is out dated.

cli/tests/upgrade-tests/main_test.go
VineethReddy02: `m.Run` will run all the test files under `upgrade-tests` in the future we might add more tests to this. So the main `install` & `upgrade` flow are global and the rest of the tests will run on top of it.

cli/tests/upgrade-tests/main_test.go
VineethReddy02: The comment line is outdated. Will update the somment line.

cli/cmd/helm/upgrade.go
VineethReddy02: I wanted this. As we need more info on failures. i.e. With `%v` or `%w` all we get is `exit status 1`. By adding the additional string output `gives us the exact output emitted by `helm`. 

I will change this to `return fmt.Errorf("failed to upgrade %s: %w", output, err)` for better logging on error.

.github/workflows/generate.yml
```diff
@@ -47,16 +47,16 @@ jobs:
...
       - name: E2E-Tests
         working-directory: ${{env.working-directory}}
         run: |
-          make e2e-tests
-
-      - name: External-DB-Tests
-        working-directory: ${{env.working-directory}}
-        run: |
-          make external-db-tests
+          export SKIP_INSTALL_TESTS=TRUE && make e2e-tests
```
cevian, line 62:
```suggestion
          SKIP_INSTALL_TESTS=TRUE make e2e-tests
```

cevian: it's not necessary to export if using only on one command

cli/cmd/helm/helmInstall.go
```diff
@@ -178,6 +194,11 @@ func (c *installSpec) installStack() error {
 		return fmt.Errorf("could not install The Observability Stack: %w \nOutput: %v", err, string(out))
 	}
 
+	if c.skipWait {
+		fmt.Println("skipping to wait for pods to come to running state. As --skip-wait is enabled.")
```
cevian, line 198:
```suggestion
		fmt.Println("skipping the wait for pods to come to a running state because --skip-wait is enabled.")
```

cli/pkg/k8s/k8s.go
```diff
@@ -509,15 +509,21 @@ func ExpandPVC(namespace, pvcName, value string) error {
...
 }
 
-func DeletePods(namespace string, labels map[string]string) error {
+func DeletePods(namespace string, labels map[string]string, forceKill bool) error {
 	pods, err := KubeGetPods(namespace, labels)
 	if err != nil {
 		return fmt.Errorf("failed to get the pods using labels %w", err)
 	}
 
+	var metaV1Options metav1.DeleteOptions
```
cevian, line 518:
```suggestion
	var deleteOptions metav1.DeleteOptions
```

cli/tests/upgrade-tests/main_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	test_utils "github.com/timescale/tobs/cli/tests/test-utils"
+)
+
+var PATH_TO_TOBS = "./../../bin/tobs"
+var PATH_TO_CHART = "./../../../chart/"
+var NAMESPACE = "ns"
+var RELEASE_NAME = "gg"
+
+const version0_2_2 = "0.2.2"
```
cevian, line 20:
```suggestion
const upgradeFromVersion = "0.2.2"
```

cli/tests/upgrade-tests/main_test.go
cevian: Hmm not quite sure that will work as, for example, TestUpgrade leaves the system after it's run in a very different state than upgradeTobsLatest(). Therefore any test that runs after TestUpgrade will not be able to rely on the state. I suspect you'll want to do this kind of setup in each test individually. But, we can leave it like this for now and see how it goes.

cevian (approved):
Some nits to fix, but overall good to go.

cli/tests/upgrade-tests/main_test.go
VineethReddy02: As the current tests. Do not need this install & upgrade to be outside. So I will move them as you suggested. Maybe in the future, I will pull them out. If I find more upgrade tescases.

PR #130, opened 2021-05-16
Add additional scrape config

VineethReddy02:
add additional scrape config to scrape all pods and services in the k8s cluster

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -104,6 +104,201 @@ kube-prometheus-stack:
               requests:
                 storage: 8Gi
 
+      # We've enabled annotation-based scraping by default for backward-compatibility
+      # and to support the largest number of use-cases out-of-the-box.
+      # We encourage people to use ServiceMonitors and PodMonitors for new components.
+      # See discussion in: https://github.com/prometheus-operator/prometheus-operator/issues/1547
+      # and more info: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prometheusioscrape
+      additionalScrapeConfigs:
+        - job_name: kubernetes-service-endpoints
+          kubernetes_sd_configs:
+            - role: endpoints
+          relabel_configs:
+            - action: keep
+              regex: true
+              source_labels:
+                - __meta_kubernetes_service_annotation_prometheus_io_scrape
+            - action: replace
+              regex: (https?)
+              source_labels:
+                - __meta_kubernetes_service_annotation_prometheus_io_scheme
+              target_label: __scheme__
+            - action: replace
+              regex: (.+)
+              source_labels:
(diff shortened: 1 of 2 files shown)


cevian (changes requested):
Where did these configs come from?

VineethReddy02:
These scrape configs are by default enabled in old Prometheus helm deployment i.e. from [here](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml#L1296)

The new Kube-Prometheus stack comes with [serviceMonitors](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L350) for Kubernetes api-server, kubelet, cadvisor and for all Kube-Prometheus components i.e. Prometheus Operator, Prometheus, Alertmanager, Kube-state-metrics, etc...

By default, the Kube-Prometheus stack doesn't scrape other workloads in the cluster for example Promscale. So added the additional scrape configs related to application workloads i.e. for pods, pods-slow, services, endpoints, and endpoints-slow. So all the workloads i.e. pods, services, endpoints running in the cluster will be scrape by these additional service discovery scrape configs if they match with required annotations/labels.

chart/values.yaml
```diff
@@ -103,6 +103,196 @@ kube-prometheus-stack:
               requests:
                 storage: 8Gi
 
+      additionalScrapeConfigs:
```
cevian, line 106:
```suggestion
      additionalScrapeConfigs:
      # We've enabled annotation-based scraping by default for backward-compatibility
      # and to support the largest number of use-cases out-of-the-box.
      # We encourage people to use ServiceMonitors and PodMonitors for new components.
      # See discussion in: https://github.com/prometheus-operator/prometheus-operator/issues/1547
      # and more info: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prometheusioscrape
```

chart/values.yaml
```diff
@@ -152,6 +347,18 @@ kube-prometheus-stack:
...
+  # serviceMonitor disable annotation based scraping
+  kube-state-metrics:
+    prometheusScrape: false
+
+  # By default node-exporter are scraped using
+  # serviceMonitor disable annotation based scraping
+  prometheus-node-exporter:
+    service:
+      annotations:
+        prometheus.io/scrape: "false"
```
cevian, line 360: do we want this in the `cli/tests/testdata/main-values.yaml ` file too?

cevian (approved):
Looks good

chart/values.yaml
VineethReddy02: Not really. As in our test we do not care about targets and the metrics that are being scraped. So all the additional config only exists in values.yaml that are shipped with default tobs helm chart.

PR #132, opened 2021-05-21
If noticed any warnings log them and parse the output

VineethReddy02:
change `combinedOut` to `output` as using combinedOut includes warnings if there are any and fails to parse operations of expected results...

This is observed commonly helm lags out warnings...

### DIFF
--- a/cli/pkg/utils/utils.go
+++ b/cli/pkg/utils/utils.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


lbogdan:
The `WARNING:` parsing feels a bit fragile, maybe a better approach would be to capture `stdout` and `stderr` separately, like for example [here](https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html#ca6597d7-e802-4d4e-9522-0e6197de19c9)?

VineethReddy02:
Thanks for the suggestion @lbogdan, I totally appreciate your input. 👍  

In the upcoming release, we have plans to switch helm dependency from CLI to helm library. So that we can drop the `helm` CLI dependency for tobs. which should also help us to ease these checks for warnings and to avoid running `helm` cmds from tobs.

cevian:
@lbogdan thanks, that change is a major improvement

cevian (approved):

PR #137, opened 2021-06-08
switch helm CLI dependency to helm library

VineethReddy02:


cli/pkg/helm/client.go
```diff
@@ -0,0 +1,557 @@
...
+func newClient(options *Options, clientGetter genericclioptions.RESTClientGetter, settings *cli.EnvSettings) (Client, error) {
+	err := setEnvSettings(options, settings)
+	if err != nil {
+		return nil, err
+	}
+
+	debugLog := options.DebugLog
+	if debugLog == nil {
+		debugLog = func(_ string, _ ...interface{}) {}
+	}
```
antekresic, line 52:
```suggestion
	debugLog := func(_ string, _ ...interface{}) {}
	if options.DebugLog != nil {
		debugLog = options.DebugLog
	}
```

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,557 @@
...
+				}
+
+				// as chart dependencies are updated fetch the chart again
+				details.helmChart, *details.chartPath, err = c.getChart(details.spec.ChartName, details.chartPathOptions)
+				if err != nil {
+					return err
+				}
+			} else {
+				return err
+			}
```
antekresic, line 275: Remove else by handling errors first.
```suggestion
			if !details.spec.DependencyUpdate {
				return err
			}
```

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,557 @@
...
+
+// chartIsInstalled checks whether a chart is already installed or not by the provided release name
+func (c *HelmClient) chartIsInstalled(release string) (bool, error) {
+	histClient := action.NewHistory(c.ActionConfig)
+	histClient.Max = 1
+	if _, err := histClient.Run(release); err == driver.ErrReleaseNotFound {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
```
antekresic, line 464:
```suggestion
	if _, err := histClient.Run(release); err != nil {
		if err == driver.ErrReleaseNotFound {
			err = nil
		}
		return false, err
	} 
```

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,557 @@
...
+	err := c.ActionConfig.Init(c.Settings.RESTClientGetter(), "", "", func(_ string, _ ...interface{}) {})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list deployed releases %v", err)
+	}
+	listClient := action.NewList(c.ActionConfig)
+	listClient.StateMask = action.ListDeployed
+	return listClient.Run()
+}
+
+func (c *HelmClient) getReleaseValues(name string, allValues bool) (map[string]interface{}, error) {
```
antekresic, line 479: Since I don't see any case that you are not getting all values, can we remove this bool flag? Its an anti-pattern to send flags like this. If there is a need to have both versions (all values and not all values), I suggest creating two functions: `getAllReleaseValues` and `getReleaseValues`.

Makes the code easier to read.

cli/pkg/helm/client_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	helmClient, err := New(opt)
+	if err != nil {
+		panic(err)
+	}
+	_ = helmClient
+}
+
+func ExampleHelmClient_AddOrUpdateChartRepo_public() {
```
antekresic, line 24: Why use snake-case naming?

https://golang.org/doc/effective_go#mixed-caps

cli/pkg/helm/helm.go
```diff
@@ -0,0 +1,86 @@
...
+func InstallUpgradeChart(chartSpec *ChartSpec) (*release.Release, error) {
+	c := getClient()
+	release, err := c.InstallOrUpgradeChart(context.Background(), chartSpec)
+	if err != nil {
+		return release, fmt.Errorf("failed to install chart %w", err)
+	}
+	return release, nil
+}
+
+func UnInstallChart(chartSpec *ChartSpec) error {
```
antekresic, line 48:
```suggestion
func UninstallChart(chartSpec *ChartSpec) error {
```

antekresic (changes requested):

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+	if c.linting {
+		err = c.lint(chartPath, values)
+		if err != nil {
+			return release, err
+		}
+	}
+
+	client.Namespace = c.Settings.Namespace()
+	// create namespace if it doesn't exist
+	client.CreateNamespace = true
```
cevian, line 236: is this the semantics we want?

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+		return err
+	}
+
+	k8sClient, err := clientset.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+
+	for _, crd := range chartInstance.CRDObjects() {
+		// use this ugly detour to parse the crdYaml to a CustomResourceDefinitions-Object because direct
```
cevian, line 385: did you get this logic from somewhere?

cevian (changes requested):
Partial review -- will finish tomorrow

cli/pkg/helm/client_getter.go
```diff
@@ -0,0 +1,32 @@
+package helm
+
+import (
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
```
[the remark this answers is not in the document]
VineethReddy02: Missed to drop them :)

cli/pkg/helm/interface.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"context"
+
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/repo"
+)
+
+//go:generate mockgen -source=interface.go -package mockhelmclient -destination=./mock/interface.go -self_package=. Client
+
+type Client interface {
```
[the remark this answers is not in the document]
VineethReddy02:
```
// Client defines the values of a helm client
type ClientInfo struct {
	Settings     *cli.EnvSettings
	Providers    getter.Providers
	storage      *repo.File
	ActionConfig *action.Configuration
	linting      bool
}
```

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+	for _, k := range helmChart.Raw {
+		if k.Name == "Chart.yaml" {
+			return k.Data, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to get Chart.yaml from the provided chart")
+}
+
+// mergeInstallOptions merges values of the provided chart to helm install options used by the client
```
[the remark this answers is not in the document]
VineethReddy02: changing the naming to `set*`

cli/pkg/helm/interface.go
```diff
@@ -0,0 +1,22 @@
+package helm
+
+import (
+	"context"
+
+	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/repo"
+)
+
+//go:generate mockgen -source=interface.go -package mockhelmclient -destination=./mock/interface.go -self_package=. Client
```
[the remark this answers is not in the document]
VineethReddy02: This is for generating mockCLient. But we do not need this dropped it.

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+
+	// set the namespace with this ugly workaround because cli.EnvSettings.namespace is private
+	// thank you helm!
+	if options.Namespace != "" {
+		pflags := pflag.NewFlagSet("", pflag.ContinueOnError)
+		settings.AddFlags(pflags)
+		err := pflags.Parse([]string{"-n", options.Namespace})
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
VineethReddy02: Dropped it.

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+
+	chartRepo.CachePath = c.Settings.RepositoryCache
+
+	_, err = chartRepo.DownloadIndexFile()
+	if err != nil {
+		return err
+	}
+
+	if c.storage.Has(entry.Name) {
+		log.Printf("WARNING: repository name %q already exists", entry.Name)
```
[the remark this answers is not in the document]
VineethReddy02: This is just a warning log if helm repo exists and we try to add/update again we log it. But I am dropping this check as we trigger `AddUpdateCHart` because we intentionally want to add/update the existing repo.

cli/pkg/helm/client.go
VineethReddy02: This to update the CRDs & this logic borrowed from the community helm-client.

cli/pkg/helm/client.go
VineethReddy02: yup, helm install creates the namespace if it doesn't exist. So we are following the same semantics...

.github/workflows/generate.yml
```diff
@@ -56,7 +56,32 @@ jobs:
         with:
           version: "v0.7.0"
 
+      - name: Upgrade-Tests
```
VineethReddy02, line 59: This file will be reverted back before merging. These changes are to make sure all the tests in CI are passing as this PR involves replace helm CLI with helm library.

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,566 @@
...
+}
+
+// install lints and installs the provided chart
+func (c *HelmClient) install(spec *ChartSpec) (*release.Release, error) {
+	release := &release.Release{}
+	client := action.NewInstall(c.ActionConfig)
+	mergeInstallOptions(spec, client)
+
+	if client.Version == "" {
+		client.Version = ">0.0.0-0"
```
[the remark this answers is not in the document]
VineethReddy02: At tobs level, we always pass the desired version to helmCLient. i.e. default to the latest version if not `--version` value provided by the user. So we can drop this version check.

cli/pkg/k8s/k8s.go
```diff
@@ -29,7 +29,7 @@ import (
 
 var HOME = os.Getenv("HOME")
 
-func kubeInit() (kubernetes.Interface, *rest.Config) {
+func KubeInit() (kubernetes.Interface, *rest.Config) {
```
VineethReddy02, line 32: Ignore this `KubeClient()` here in the upcoming PR `kubeclient` will be created in the tobs cmd level & will be passed to the `k8s` package.

cli/pkg/helm/client.go
```diff
@@ -0,0 +1,509 @@
...
+	err := os.Setenv("HELM_NAMESPACE", namespace)
+	if err != nil {
+		log.Fatalf("failed to set HELM_NAMESPACE env variable %v", err)
+	}
+}
+
+// after install/upgrade we need to unset helm namespace set by us
+// to make sure we don't leave unwanted envs set.
+func unSetHelmNamespaceEnv() {
+	err := os.Setenv("HELM_NAMESPACE", "")
```
cevian, line 96: are we sure the old value was ""? Shouldn't we store the old value so we can restore correctly?

cli/pkg/helm/client.go
cevian: could you add a comment with the link to where the code is borrowed from?

cli/pkg/helm/helper.go
```diff
@@ -0,0 +1,180 @@
...
+	Namespace  string `json:"namespace"`
+	Revision   string `json:"revision"`
+	Updated    string `json:"updated"`
+	Status     string `json:"status"`
+	Chart      string `json:"chart"`
+	AppVersion string `json:"app_version"`
+	Version    string
+}
+
+func (c *clientImpl) GetChartMetadata(chart string) (*ChartMetadata, error) {
```
cevian, line 37: all methods of clientImpl should go into client.go

cli/pkg/helm/values.go
```diff
@@ -0,0 +1,131 @@
+package helm
```
cevian, line 1: can we add some tests for these functions?

cli/tests/helm-tests/helm_test.go
```diff
@@ -0,0 +1,70 @@
+package helm_tests
```
cevian, line 1: Glad to see these tests. 1 comment: can we test for more than just lack of errors? Like verify something more than just err!=nil

cevian (changes requested):
Much better. A few changes requested. Would be great to see some more tests too

cevian (commented):
Looks great. Don't forget to undo the ci/cd changes you wanted to roll back before merge

cevian (approved):
Looks great. Don't forget to undo the ci/cd changes you wanted to roll back before merge

PR #144, opened 2021-07-07
fix version in helm tests

VineethReddy02:


cevian (approved):
Looks good. One small comment

PR #148, opened 2021-07-13
fix tobs uninstall flow

VineethReddy02:
1. Removed the duplicate code
2. Added additional secret deletion which isn't deleted on performing helm chart uninstall.


cevian (approved):
LGTM

PR #161, opened 2021-08-10
Minor bug fixes

VineethReddy02:
1. Add default Prometheus instance retention to `1d`. 
2. Change Promscale datasource name from Prometheus to Promscale. This will be used to display the name in Grafana UI. 
3. Tobs install should create secrets for versions >= `0.3.0`. Added the = which was missed. This is a bug only if a user installs the tobs by explicitly providing the version as `0.3.0`. **Note**: Installation based on version wasn't available in version 0.3.0. So this just makes the validation better. 
4. If the Kubernetes cluster is big with 10's of helm releases finding the existing tobs installation was failing as the helm client was timed out. So now to find the existing tobs helm release we will search in a specific namespace which makes finding the release much easier. 
5. For some reason if an upgrade is failed or went into a pending state. Tobs CLI couldn't find the release as it isn't in the deployed state. This makes debugging a failed upgrade tougher and CLI can't help in such cases. So from now tobs CLI will list the failed and pending releases to retry the upgrades or find the current state of a release. 
6. On upgrade attempt to apply CRDs creation only if we notice version change between deployed version vs new version. Currently, CRDs creation message misleading on upgrading to the same version with changes in `values.yaml`.  
7. In daily run tests HA tests we are experiencing flakiness, so increasing the retry mechanism sleep overtime to 5 seconds from 3 seconds. This gives a decent amount of time for Promscale HA leader transfer to happen.

### DIFF
--- a/cli/cmd/upgrade/upgrade.go
+++ b/cli/cmd/upgrade/upgrade.go
@@ -105,7 +105,7 @@ func upgradeTobs(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	deployedChart, err := helmClient.GetDeployedChartMetadata(root.HelmReleaseName)
+	deployedChart, err := helmClient.GetDeployedChartMetadata(root.HelmReleaseName, root.Namespace)
 	if err != nil {
 		if err.Error() != utils.ErrorTobsDeploymentNotFound(root.HelmReleaseName).Error() {
 			return err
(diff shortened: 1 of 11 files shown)


cli/cmd/upgrade/upgrade.go
```diff
@@ -232,7 +232,10 @@ func (c *upgradeSpec) UpgradePathBasedOnVersion() error {
 		return fmt.Errorf("failed to parse 0.2.2 version %w", err)
 	}
 
-	if nVersion >= version0_4_0 {
+	// kube-prometheus is introduced on tobs >= 0.4.0 release
+	// so create CRDs if version >= 0.4.0 and only create CRDs
+	// if version change is noticed in upgrades...
+	if nVersion >= version0_4_0 && nVersion != dVersion {
```
cevian, line 238: hmm so CRDs will be re-created every time? Then the crd urls on line 470 that have "v0.47.0" encoded in the URL are probably not appropriate. Thats probably a bug we should fix in this pr

cevian (changes requested):
Everything looks good except for the CRD bug below

cli/cmd/upgrade/upgrade.go
cevian: Oh nevermind I see how that's supposed to work

cevian (approved):

cevian (approved):
Good work

PR #162, opened 2021-08-12
Bump kube-prometheus-stack for multi-cluster support

VineethReddy02:


cli/cmd/upgrade/upgrade.go
```diff
@@ -467,14 +467,14 @@ func parsePgBackRestConf(data string) map[string]string {
...
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_podmonitors.yaml",
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_probes.yaml",
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml",
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml",
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_thanosrulers.yaml",
-	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.47.0/example/prometheus-operator-crd/monitoring.coreos.com_prometheusrules.yaml",
+	"https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.49.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml",
```
cevian, line 470: is there anyway we can have a test to make sure this version is aligned with the kube-prometheus-stack version? Otherwise this is very easy to forget to do.

chart/docs/values-config.md
```diff
@@ -0,0 +1,128 @@
+# Tobs helm values config
```
cevian, line 1: hmm. I think there are two different types of config here.
1. config which you need only for things that don't use the CLI (high-availability)
2. config applicable to both helm chart and CLI. (multicluster)

Type 1 is ok here. I'd move type 2 stuff into a new top-level /docs directory. It does not belong under /chart and I'd add new links from the front-page readme here.

chart/docs/values-config.md
```diff
@@ -0,0 +1,128 @@
...
+* Configure Prometheus remote-write to Promscale loadbalancer service in data aggregation cluster and add the external label representing the current cluster name to differentiate the visualisation on per cluster basis in Grafana. 
+
+```
+kube-prometheus-stack:
+  prometheus:
+    prometheusSpec:
+      externalLabels:
+        cluster: <clusterName>  
+      remoteRead:
+      - url: "<PROMSCALE_ENDPOINT>/read"
```
cevian, line 124:
```suggestion
      - url: "<PROMSCALE_SERVICE_ENDPOINT_OF_DATA_AGGREGATION_CLUSTER>/read"
```

chart/docs/values-config.md
```diff
@@ -0,0 +1,128 @@
...
+```
+kube-prometheus-stack:
+  prometheus:
+    prometheusSpec:
+      replicaExternalLabelName: "__replica__"
+      prometheusExternalLabelName: "cluster"
+      replicas: 3
+```
+
+## Multi-cluster support
```
cevian, line 43: can we have a diagram for this section showing the various clusters etc.

cevian (approved):
Some optional suggestions but looks good to merge

cli/cmd/upgrade/upgrade.go
VineethReddy02: Tests definitely sound reasonable. But usually, during the kube-prometheus version bump, we read the upgrade guide in kube-prometheus helm chart to understand the changes for upgrades. 

I will add the tests on this.

chart/docs/values-config.md
VineethReddy02: So currently using CLI we offer only abstract features with flags like `HA`, `integration with forge`, etc... all these features can also be configured within `values.yaml`. All the flexibility lies in the `values.yaml` so the intention behind creating this file is to document all possible configurations across different components in tobs.

Yes, I agree with you only config related to `values.yaml` will reside in this file. Anything about CLI will be in a new file in the high level root directory of this repo.

chart/docs/values-config.md
```diff
@@ -0,0 +1,130 @@
...
+
+<img src="./../../docs/assets/multi-cluster.png" alt="multi-cluster setup diagram" width="800"/>
```
cevian, line 45: I believe you forgot to add this image in the PR?

cevian (approved):
Looks good except for missing image

chart/docs/values-config.md
VineethReddy02: Added it to the PR here is the file: https://github.com/timescale/tobs/pull/162/files#diff-1a7cd77999e454c5a336fcb163eb5108d70057981aa40d648c619671eba20b3e

Issue #169, opened 2021-09-22
tobs install failure on gcp 


dnck:
With the kubeconfig 

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: OMIT 
    server: OMIT
  name: OMIT
contexts:
- context:
    cluster: OMIT
    namespace: OMIT
    user: OMIT
  name: OMIT
current-context: OMIT
kind: Config
preferences: {}
users:
- name: OMIT
  user:
    auth-provider:
      config:
        access-token: OMIT
        cmd-args: config config-helper --format=json
        cmd-path: gcloud
        expiry: "2021-09-22T14:53:32Z"
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```

Running, 

```
./tobs install --chart-reference ../chart
```

from the `cli` directory fails, 

```
... no Auth Provider found for name "gcp"
```

This seems to be a regression to https://github.com/timescale/tobs/issues/68.

avinava-mycelium:
Hello All,
Thanks for this awesome implementation.
This will surely make monitoring way easier.

However I am still facing the above-mentioned issue and when I check the code, I can see that the code changes are yet to be merged?

In that case, can we know when will this release be done?

Additionally, I have to run the CLI by going to the base folder as the symlink creation fails in MAC..?

PR #172, opened 2021-09-23
Add blank import for k8s client auth in cli for gcp support

dnck:
For bug report #169, running the `init()` function of `k8s.io/client-go/plugin/pkg/client/auth` in `./cli/pkg/client_impl.go` seems to fix the issue. 

According to the docs, I should test with the `cli/tests/e2e-tests.sh`, but current issues with posix sh

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -3,7 +3,31 @@ module github.com/timescale/tobs/cli
 go 1.17
 
 require (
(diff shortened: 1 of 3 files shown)


VineethReddy02:
@dnck Thanks for the contribution. Can you sign the CLA (Contributor License Agreement)?

Issue #181, opened 2021-10-08
Helm fails when deploying a remote observer cluster


SteveMaddison:
Following the steps [here](https://github.com/timescale/tobs/blob/master/chart/docs/values-config.md#deploying-the-observer-cluster) to deploy an observer cluster, Helm is unable to render `NOTES.txt` due to `template "timescaledb.fullname"` (lines 123, 126) and `template "clusterName"` (lines 140, 146) not being defined. I suspect the entire section on TimescaleDB isn't relevant could actually be if-ed out in this situation.

Sanitized values to reproduce:
```
timescaledb-single:
  enabled: false
promscale:
  enabled: false
kube-prometheus-stack:
  grafana:
    enabled: false
kube-prometheus-stack:
  prometheus:
    prometheusSpec:
      externalLabels:
        cluster: mycluster  
      remoteRead:
      - url: http://promscale.example.com/read
        readRecent: true
      remoteWrite:
      - url: http://promscale.example.com/write
```

VineethReddy02:
Hi @SteveMaddison 
Thanks for reporting this issue. This issue is only happening when the tobs helm chart is used without using the tobe CLI. 

I will create a PR with the fix.

PR #187, opened 2021-10-20
Fix disable TimescaleDB issue 

VineethReddy02:
Currently, If TimescaleDB is disabled. The installation can work in two modes
1. Connecting to an external TimescaleDB
2. Deploying an observer cluster i.e. only Prometheus components will be deployed (this setup acts as an observer cluster) will write the data to a centralized Promscale instance outside the cluster. 

Issues fixed:
1. If TimescaleDB isn't deployed. We by default consider that we are connecting to an external database. But there is another case where the tobs is deployed as the observer cluster. In this case, we do not need any DB secrets to be created + the rendering of `NOTES.txt` needs to be fixed.

Fixes: #181

### DIFF
--- a/chart/templates/NOTES.txt
+++ b/chart/templates/NOTES.txt
@@ -7,7 +7,7 @@
 {{- end }}
 {{- if index .Values "timescaledb-single" "enabled" }}
 🐯 TimescaleDB
-{{- else }}
+{{- else if index .Values "timescaledbExternal" "enabled" }}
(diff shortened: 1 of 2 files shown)


kevinlmadison:
Do we know when the next release will be? I don't know how to work around the issue that this fixes (#188)

VineethReddy02:
@kevinlmadison the release will be out tomorrow.

PR #207, opened 2022-01-14
Update Jaeger-Promscale to upstream Jaeger

VineethReddy02:


chart/templates/jaeger-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+{{ if .Values.opentelemetryOperator.enabled -}}
```
VineethReddy02, line 1: This is the same old file like the one [below (shown as deleted)](https://github.com/timescale/tobs/pull/207/files#diff-244f8b88267bbd39adc33961fa72236162f87c283780c2a3c5bb96e7fd45d24f )

With few changes as dropping volumes mounts from configmap to passing them as the args.

chart/values.yaml
```diff
@@ -446,15 +446,13 @@ opentelemetryOperator:
             exporters: [logging, otlp]
             processors: [batch]
 
-  jaegerPromscaleQuery:
-    image: timescale/jaeger-query-proxy:0.0.1
-    endpoint: "{{ .Release.Name }}-jaeger-promscale.{{ .Release.Namespace }}.svc.cluster.local:16686"
+  jaeger:
+    image: jaegertracing/jaeger-query:1.30
+    endpoint: "{{ .Release.Name }}-promscale-connector.{{ .Release.Namespace }}.svc.cluster.local:9202"
```
cevian, line 451: This line reads a bit weird now, because it's not the endpoint of the jaeger query image above. It's only really used by Grafana. So maybe move it up into the Grafana section? An example of the awkwardness: what if you wanted to disable the jaeger query ui but keep the Grafana datasource setup.

cevian (changes requested):
Overall looks good but one UX change requested

chart/values.yaml
```diff
@@ -446,15 +448,12 @@ opentelemetryOperator:
             exporters: [logging, otlp]
             processors: [batch]
 
-  jaegerPromscaleQuery:
-    image: timescale/jaeger-query-proxy:0.0.1
-    endpoint: "{{ .Release.Name }}-jaeger-promscale.{{ .Release.Namespace }}.svc.cluster.local:16686"
+  jaeger:
+    image: jaegertracing/jaeger-query:1.30
```
cevian, line 452: nit: I might consider making the env variables (SPAN_STORAGE_TYPE=grpc-plugin) also configurable here. But that is a minor point and shouldn't block the PR

cevian (approved):

chart/values.yaml
VineethReddy02: Updated!

paulfantom (approved):
:shipit:

PR #217, opened 2022-01-19
Tobs telemetry 

VineethReddy02:


paulfantom (changes requested):
The same is done in #215

PR #285, opened 2022-03-09
Docs: Remove typo

politician:
The flag `-n <name>` is not needed in the docs


cevian:
Asking @VineethReddy02  and @paulfantom to review instead

paulfantom (approved):

PR #583, opened 2022-09-15
add avalanche to tobs, update kind

nhudson:

- fixes #

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -15,7 +15,7 @@ keywords:
   - tracing
   - opentelemetry
 
-version: 14.9.0
+version: 15.0.0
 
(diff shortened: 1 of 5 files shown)


paulfantom (changes requested):
Avalanche shouldn't be added to tobs. We should add it only when doing benchmarking.

paulfantom (changes requested):
The following configuration is needed:
- prometheus deployed in 1 replica
- prometheus not sending data to promscale (remote_read and remote_write configs empty)
- alertmanager NOT deployed
- tracing disabled
- avalanche deployed separately. Can be joined in a higher-level helm chart.

All this needs to happen outside of tobs helm chart (or in some `contrib/` directory).

PR #645, opened 2022-11-14
Set stats_temp_directory to tmp directory

alejandrodnm:

Moves the stats temporary file outside the data directory to improve
disk performance.

According to the docs:

```
stats_temp_directory (string)

Sets the directory to store temporary statistics data in. This can be
a path relative to the data directory or an absolute path. The default
is pg_stat_tmp. Pointing this at a RAM-based file system will decrease
physical I/O requirements and can lead to improved performance. This
parameter can only be set in the postgresql.conf file or on the server
command line.
```

PG15 rewrote the statistics collector, from that version onward the
statistics are no longer stored in files and the `stats_temp_directory`
config options was deprecated. Once the DB upgrades to PG15 this setting
can be removed.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -12,7 +12,7 @@ keywords:
   - monitoring
   - tracing
   - opentelemetry
-version: 17.16.0
+version: 17.17.0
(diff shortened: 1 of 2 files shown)


onprem (approved):
LGTM, thanks!

